PR #5, opened 2017-03-15
Complexity collector

JiriPavela:
New addition to the perun collectors set. 

Unable to fully integrate the module due to missing command line args, but the collector itself is functional (tested the module after the integration). See the 'complexity.py' test run (commented out at the bottom of the file) which uses the config dictionary as a argument list.


tfiedor:
1) Pull Request contains broken imports of custom packages. Fix this issue either by using the full qualified name everywhere (import perun.collect.complexity.makefiles) or at least partially qualified name (i.e. import perun.collect.complexity.makefiles as makefiles)

2) I made a little hack to circumvent the config, however, it still is not enough and I cannot test it more, since there seems to be missing CMakeFiles.txt in the target directory :).

So please solve these two things, push them to this pull request, I will test it again and merge if everything is OK.

JiriPavela:
Hopefully fixed the integration. The paths are corrected and the test run is updated, should be now functional from the top perun directory. I'm not sure if you need the test configuration and test run, so it is commented out at the bottom of the complexity.py, uncomment if nedded :).

Also renamed and documented the profiling api, collector progress is now printed to the console.

tfiedor:
Good thing is I managed to run it and produce profiles.

However, there are still few minor issues, that needs to be patched and that I had to circumvent in my clone:
1) Please apply the following patch ([setup.txt](https://github.com/tfiedor/perun/files/849957/setup.txt)). Perun should really be run from anywhere and should be installed inside the local site packages (or somewhere else) and this patch will ensure that all of the necessary files will be present for the complexity collector (it is temporal workaround though, I'm not sure if everything is needed)
2) Statement at [complexity.py#L128](https://github.com/JiriPavela/perun/blob/17511f8cd1edee75550fb9a2e900cf149a45df43/perun/collect/complexity/complexity.py#L128) generates incorrect profile format. It should be of the following form:
`
kwargs['profile'] = {'global' : {'timestamp' : ..., 'resources': resources}}}
`
The `timestamp` (overall run of the profiled program) should be inferable from the collected data; and `resources` is the list that you previously returned. And don't forget to format the above line nicely ;).

After these two things are fixed, I should have no further issues with this PR.

JiriPavela:
Applied the setup patch and updated the after function, which should now provide the correct output profile format :).

Issue #9, opened 2017-05-27
Zero value in address records.


rpodola:
Rarely there are zero values in the address records in profile which seems to be invalid.

tfiedor:
Do you have any idea, why such records appear in profiles? AFAIK, zero addresses should be protected, as this is, where kernel space is.

rpodola:
There are _ptr_ pointer recorded instead of _nptr_ in code bellow (malloc.c)
`
   if(!profiling && ptr != NULL){
      
      ad_log("realloc", size, nptr);

      if(ptr != nptr){
         ad_log("free", 0, ptr);
      }
`

tfiedor:
Closing the issue, as this was fixed by PR #12.

PR #11, opened 2017-05-28
Zero value address fix

rpodola:
This will fix the issue #9 with zero value addresses when reallocation using realloc() was produce this type of invalid record. Should be able to merge now.


tfiedor:
Still not OK. Looks like something went wrong with the merge in your branch, because after review of overall changes of this commit, some of the code chunks would be "reintroduced" and the code would stop working (as demonstrated by failing travis). So I cannot allow merging this.

Since this should be just a merge of c7186a9 commit, it should contain just that commit. I think easiest solution would be checking out fresh develop branch from upstream, then checkout fix branch and cherry pick or patch the changes of c7186a9 into the newly checkouted fix branch. Or send me the patch through email and I will do it manually in my repo and push the fix.

rpodola:
Looks like a pulling develop branch from upstream a several times just messed up whole branch and add some code duplicities. I will refresh whole branch then with reset --hard and keep my fingers crossed.

PR #19, opened 2017-07-02
Add Flame graph visualization.

rpodola:
Add Flame graph visualization using 3rd party visualizer by Brendan Gregg. There is known issue, when installation of the Perun does not copy visualizer (written in Perl) to installation folder. Visualization is working after manual copying the visualizer into `perun/view/flamegraph/`, although it would be convenient to edit the installation setup of the Perun.


tfiedor:
Awesome! Thanks again.

Codacy/Codeclimate issues seems minor, so no need to bust your head over them. I will try it later out (after I'm done with current feature branch) and probably make minor changes (add test, style, etc.).

PR #22, opened 2017-07-17
Update the complexity collector cli (add new parameter, unify the nam…

JiriPavela:
…es) and implement the cli using click.

circ.rst: update the documentation for internal configuration file
configurator.py: add the cli options
Makefile: update the makefile to build only the libraries
configuration.cpp, configuration.h: rename / add the parameters for circ, fixes
profile.cpp, profile.h: fixes
run.py: add the click parameters
libprofapi.so, libprofile.so: updated versions of the library


perun/collect/complexity/configurator.py
```diff
@@ -97,14 +102,23 @@ def _config_write_config(config_handle, executable_path, runtime_filter, include
...
+    if 'internal_data_filename' not in config:
+        config['internal_data_filename'] = DEFAULT_DATA_FILENAME
+    if 'internal_storage_size' not in config:
+        config['internal_storage_size'] = DEFAULT_STORAGE_SIZE
+    if 'internal_direct_output' not in config:
+        config['internal_direct_output'] = DEFAULT_DIRECT_OUTPUT
+
+    # Append the internal configuration
+    conf = {
+        'internal_data_filename': config['internal_data_filename'],
```
tfiedor: I would suggest to change this to the following:
`'internal_data_filename': config.get('internal_data_file', DEFAULT_DATA_FILENAME)`
and then remove the lines 106-107. 

And then similar change for lines 116-117. This is IMO cleaner solution, and, moreover, it will remove the misses from codecov and hence, the whole PR should get ✔️ and will be ready for a merge.

Also consider, if you really need two separate configuration dictionaries (config and conf—which is further passed).

JiriPavela: Updated the dict initialization code and changed the identifiers to better reflect the meaning of both dictionaries. The dictionary passed to the method contains more keys (collector job data) and thus new configuration dictionary is created (the internal collector configuration with less keys).

tfiedor: Nice job, all green 👍

PR #25, opened 2017-08-01
Refactored and integrated regression analysis module merge.

JiriPavela:
Pull request contains refactored and integrated regression analysis module.

The directory example in regression_analysis module contains two examples of resulting profiles after the postprocess is complete.


perun/postprocess/regression_analysis/example/full_computation.perf
```diff
@@ -0,0 +1,322 @@
...
+      "mixed(time delta)": "ms"
+    },
+    "params": ""
+  },
+  "postprocessors": [],
+  "collector_info": {
+    "name": "complexity",
+    "params": {
+      "sampling": [
+        [
```
tfiedor: Can you elaborate, what is included in this list? I was wondering whether dictionary with named params would be a little bit more verbose. I mean, one of our goals is human readable format of profiles, so we should eventually get there :)

perun/postprocess/regression_analysis/example/full_computation.perf
```diff
@@ -0,0 +1,322 @@
...
+        "type": "mixed",
+        "structure-unit-size": 0,
+        "uid": "SLListcls::~SLListcls()"
+      }
+    ],
+    "time": "0.000165s",
+    "regression_analysis": [
+      {
+        "r_square": 0.08716810077128712,
+        "x_limits": [
```
tfiedor: Again, can you shortly elaborate this field? What is included in this list? Are these representation of intervals? If this has known fixed length, I would again consider using dictionary with named fields instead.

perun/postprocess/regression_analysis/example/full_computation.perf
```diff
@@ -0,0 +1,322 @@
...
+          0,
+          9
+        ],
+        "model": "exponential",
+        "method": "full",
+        "uid": "SLList_insert(SLList*, int)",
+        "coeffs": [
+          0.962897772531164,
```
tfiedor: These coefficients are b0 and b1, right? The same as above...would it be better for us to transform it to nested dictionary?

perun/postprocess/regression_analysis/example/full_computation.perf
```diff
@@ -0,0 +1,322 @@
...
+        ]
+      },
+      {
+        "r_square": 0.09818181818181748,
+        "x_limits": [
+          0,
+          9
+        ],
+        "model": "linear",
+        "method": "full",
```
tfiedor: Is this important in future (e.g. when one does visualization)? I'm currently wondering what asset it has to leave it in profile. Do you have some ideas for it?

perun/postprocess/regression_analysis/example/full_computation.perf
```diff
@@ -0,0 +1,322 @@
...
+      {
+        "amount": 2,
+        "subtype": "time delta",
+        "type": "mixed",
+        "structure-unit-size": 0,
+        "uid": "SLListcls::~SLListcls()"
+      }
+    ],
+    "time": "0.000165s",
+    "regression_analysis": [
```
tfiedor: Just like we talked yesterday..."models" would probably be more suitable name.

perun/postprocess/regression_analysis/example/full_computation.perf
JiriPavela: Addressed in pull request #26 

changed to dictionary with more explicit keys.

perun/postprocess/regression_analysis/example/full_computation.perf
JiriPavela: Changed to x_interval_start and x_interval_end.

perun/postprocess/regression_analysis/example/full_computation.perf
JiriPavela: Changed the list members to more explicit dictionaries.

perun/postprocess/regression_analysis/example/full_computation.perf
JiriPavela: As mentioned on jabber, this will be useful in visualisation titles and file names.

perun/postprocess/regression_analysis/example/full_computation.perf
JiriPavela: Changed.

JiriPavela:
Addressed pylint, codacy and codeclimate issues. Two new test functions created for correct and incorrect cli inputs.

Data provider changes will be in separate branch/pull request to allow more experiments.

tfiedor:
Greato! Thanks for addressing the issues and my remarks. Everything is green and we are (almost) ready for the great merge. I will do one more through code review either today or tomorrow and if I don't find anything major, I will merge the stuff.

Nice job. 👍

perun/collect/complexity/symbols.py
```diff
@@ -231,7 +231,10 @@ def _dismantle_symbols(symbol_map):
     specification_map = dict()
     for key in symbol_map.keys():
         # Process each symbol from the map
-        specification_map[key] = _process_symbol(symbol_map[key])
+        try:
+            specification_map[key] = _process_symbol(symbol_map[key])
+        except exceptions.UnexpectedPrototypeSyntaxError:
+            pass
```
tfiedor: You sure it is good to catch this exception and ignore it? Maybe at least issue some warning?

perun/postprocess/regression_analysis/methods.py
```diff
@@ -0,0 +1,445 @@
...
+
+    for chunk in data_gen:
+        try:
+            for result in _METHODS[method](chunk[0], chunk[1], models, **kwargs):
+                result['uid'] = chunk[2]
+                result['method'] = method
+                result = _transform_to_output_data(result, ['uid', 'method'])
+                analysis.append(result)
+        except exceptions.GenericRegressionExceptionBase as e:
+            print("info: unable to perform regression analysis on function '{0}'.".format(chunk[2]))
```
tfiedor: This should rather be a warning (from perun/utils/log) so we can filter these stuff out in the future.

perun/postprocess/regression_analysis/methods.py
```diff
@@ -0,0 +1,445 @@
...
+            print("  - " + str(e))
+    return analysis
+
+
+def full_computation(x_pts, y_pts, computation_models, **kwargs):
+    """The full computation method which fully computes every specified regression model.
+
+    The method might have performance issues in case of too many models or data points.
```
tfiedor: Note to ourselves: Maybe we could speed this up? We need to stress test this.

perun/postprocess/regression_analysis/methods.py
```diff
@@ -0,0 +1,445 @@
...
+    Returns:
+        the generator object which produces computed models one by one for every interval as
+        a transformed output data dictionary
+
+    """
+    # Sort the regression data
+    x_pts, y_pts = tools.sort_points(x_pts, y_pts)
+    # Split the data into intervals and do a full computation on each one of them
+    for part_start, part_end in tools.split_sequence(len(x_pts), kwargs['steps']):
+        interval_gen = full_computation(x_pts[part_start:part_end], y_pts[part_start:part_end],
```
tfiedor: Does this need to be full_computation? The functions should have the same API, right? So could full_computation be a parameter to this function? The easiest would be to make full_computation as default parameter.

perun/postprocess/regression_analysis/regression_models.py
```diff
@@ -0,0 +1,130 @@
...
+            if model not in _MODELS.keys():
+                raise exceptions.InvalidModelException(model)
+            else:
+                yield _MODELS[model].copy()
+
+# Supported models properties
+# Each model record contains the parameters required by the computational functions,
+# the data generator and list of functions.
+# The record can also contain optional parameters as needed.
+_MODELS = {
```
tfiedor: By the way, we were talking about constant model. Will it be easy to include it in short time?

PR #26, opened 2017-08-25
Complexity collector codestandard fix

JiriPavela:
Fixed various issues from pylint in complexity collector module and changed sampling data structure in profile.


perun/collect/complexity/symbols.py
```diff
@@ -174,11 +177,10 @@ def _get_symbols(executable_path, columns):
     # Convert the columns list to string parameter
     columns_str = ','.join(str(column) for column in columns)
 
-    # # readelf -sW exec | awk '$4 == "FUNC" && $7 != "UND"' | awk '{$2=$2};1' | cut -d' ' -f columns
-    cmd = "readelf -sW " + executable_path + \
-          " | awk '$4 == \"FUNC\" && $7 != \"UND\"' | awk '{$2=$2};1' | cut -d' ' -f " + columns_str
-    ps = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
-    symbols = ps.communicate()[0]
+    cmd = ("readelf -sW " + executable_path + " | awk '$4 == \"FUNC\" && $7 != \"UND\"' | "
```
tfiedor: I am wondering, how many times is this subprocess called? Just once?

Because, when I was playing with memory-collector I ran into huge performance degradation because of subsequent calls to addr2line and had to cache the results to obtain reasonable times.

tfiedor:
Don't worry about the failing codecov, I will fix it later myself (should be quick fix for me in my tests). I reviewed the changes and they seem fine, so this is (almost) ready for a merge.

perun/collect/complexity/run.py
```diff
@@ -189,30 +189,54 @@ def _process_file_record(record, call_stack, resources, address_map):
...
+        param(object): the parameter object
+        value(list): the list of sampling values
+    Returns:
+        list of dict: list of sampling dictionaries
+    """
+    if value is not None:
+        # Initialize
+        sampling_list = []
+        # Transform the tuple to more human readable dictionary
+        for sample in value:
```
tfiedor: Note: List comprehension would work as well in this case.

PR #29, opened 2017-09-24
Add extensions + transformations to regression analysis module

JiriPavela:
Extended the regression analysis with transformations - an easily extendable module capable of performing various modification of 'models' results, e.g. conversion from coefficients to points that can be plotted.

### DIFF
--- a/perun/postprocess/regression_analysis/extensions/plot_models.py
+++ b/perun/postprocess/regression_analysis/extensions/plot_models.py
@@ -1,0 +1,147 @@
+""" Extension for regression model coefficients transformation into array of points. The points
+    array can be then used for model plotting as a series of lines forming a curved line.
+"""
+
+import perun.postprocess.regression_analysis.tools as tools
+import numpy as np
+
+
+# Default model curve smoothness specified as number of points generated from x interval
+# The higher, the smoother are the curves
+DEFAULT_SMOOTHNESS = 51
+
+
+def model_plot_computation(data):
+    """ The model plotting computation wrapper. Handles common operations regardless of model,
+        uses 'y_pts_func' to compute y values, which can be model-specific
+
+    Expects 'y_pts_func', 'x_interval_start', 'x_interval_end', 'b0' and 'b1' keys in the
+    data dictionary.
+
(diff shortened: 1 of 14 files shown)


perun/postprocess/regression_analysis/extensions/plot_models.py
```diff
@@ -0,0 +1,147 @@
...
+    array can be then used for model plotting as a series of lines forming a curved line.
+"""
+
+import perun.postprocess.regression_analysis.tools as tools
+import numpy as np
+
+
+# Default model curve smoothness specified as number of points generated from x interval
+# The higher, the smoother are the curves
+DEFAULT_SMOOTHNESS = 51
```
tfiedor: Minor Remark: Maybe explaining, why 51 was chosen constant would increase the readability and help others when adjusting the smootness.

perun/postprocess/regression_analysis/extensions/plot_models.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def model_plot_computation(data):
+    """ The model plotting computation wrapper. Handles common operations regardless of model,
+        uses 'y_pts_func' to compute y values, which can be model-specific
+
+    Expects 'y_pts_func', 'x_interval_start', 'x_interval_end', 'b0' and 'b1' keys in the
```
tfiedor: y_pts_func is f(x) that is used to compute the y value, right? Maybe we could name it just "f_x" or something like this, or is it used somewhere already?

perun/postprocess/regression_analysis/tools.py
```diff
@@ -170,4 +173,24 @@ def r_square_safe(sse, sst):
     try:
         return 1 - sse / sst
     except ZeroDivisionError:
-        return _R_SQUARE_DEFAULT
+        return R_SQUARE_DEFAULT
+
+
+def linspace_safe(start, end, steps):
```
tfiedor: This name is hard to say naturally. Can we enhance the name to something more verbose? split_space_safely? Or something similar?

PR #30, opened 2017-09-28
Scatter plot (beta version)

JiriPavela:
This pull request will add integrated scatter plot view module. The scatter plot is currently in beta version and does not provide complete functionality yet. Nevertheless it is still capable of basic plotting of complexity profiles with or without regression models.

The beta version contains some workarounds and temporary solutions, which will be removed in the full version (further discussion about certain decisions/module design is needed).


perun/view/scatter_plot/run.py
```diff
@@ -0,0 +1,115 @@
...
+                    `                       /o       ```````````````````
+                    -                     _/         `  o o = <points> `
+                    `                   _- o         `    _             `
+    <of>            -               __--o            `  _-  = <models> `
+                    `    _______--o- o               `                 `
+                    -    o  o  o                     ```````````````````
+                    `
+                    +````||````||````||````||````
+
+                                <per>
```
tfiedor: Nice job, that you fully preserver the style of the rest :-) 👍

perun/view/scatter_plot/run.py
```diff
@@ -0,0 +1,115 @@
...
+        return "Plot of '{}' per '{}'".format(ctx.params['of_key'], ctx.params['per_key'])
+    return value
+
+
+@click.command()
+@click.option('--of', '-o', 'of_key', default='amount', nargs=1,
+              show_default=True, type=str,
+              is_eager=True, callback=cli_helpers.process_resource_key_param,
+              help="Data source for the scatter plot, i.e. what will be displayed on Y axis.")
+@click.option('--per', '-p', 'per_key', default='structure-unit-size', nargs=1,
```
tfiedor: Currently, the default of "structure-unit-size" is great since currently we mostly use the scatter plot for dependencies between structure size and time. We should think of future, though, where the "safest" default would be snapshot or I am thinking of something called "order" (or "class"), which would be metric-independent (maybe I told you about it---I was thinking about simply ordering the values, and clustering the near points to same class).

perun/view/scatter_plot/run.py
tfiedor: I would keep it short, just "scatterplot" the "_" is annoying to type.

perun/view/scatter_plot/run.py
```diff
@@ -0,0 +1,115 @@
...
+              help="Label on the X axis of the scatter plot.")
+@click.option('--y-axis-label', '-yl', metavar="<text>", default=None,
+              callback=cli_helpers.process_bokeh_axis_title,
+              help="Label on the Y axis of the scatter plot.")
+@click.option('--graph-title', '-gt', metavar="<text>", default=None, callback=process_title,
+              help="Title of the scatter plot.")
+@click.option('--view-in-browser', '-v', default=False, is_flag=True,
+              help="Will show the graph in browser.")
+@pass_profile
+def scatter_plot(profile, filename, view_in_browser, **kwargs):
```
tfiedor: I would keep it short, just "scatterplot" the "_" is annoying to type.

perun/view/scatter_plot/factory.py
```diff
@@ -0,0 +1,156 @@
...
+        profile(dict): loaded perun profile
+
+    Returns:
+        generator: generator: resources and models slices of unique uid as pair
+            data_slice(pandas.DataFrame)
+            uid_models(list)
+    """
+    # Get resources for scatter plot points and models for curves
+    resource_table = converters.resources_to_pandas_dataframe(profile)
+    models = [m[1] for m in list(query.all_models_of(profile))]
```
tfiedor: Mm, maybe list(map(itemgetter(1), querry.all_models_of(profile)))*?

(*) Not sure if this is syntactically correct though.

perun/view/scatter_plot/factory.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Arguments:
+        graph(charts.Graph): the scatter plot without models
+        models(list): list of models to plot
+
+    Returns:
+        charts.Graph: the modified graph with model curves renderers
+    """
+    # Get unique colors for the model curves
+    colour_palette = palettes.viridis(len(models))
```
tfiedor: Why not use colour generator from bokeh_helpers?

perun/view/scatter_plot/run.py
JiriPavela: Changed to 'scatter' - shorter and preserves the meaning of the graph type.

perun/view/scatter_plot/factory.py
JiriPavela: Changed.

perun/view/scatter_plot/factory.py
JiriPavela: This would require to transform the 'models' into pandas dataframe - right now the 'models' is a simple list of model values. The transformation will be solved in the full version eventually.

perun/view/scatter_plot/run.py
JiriPavela: Agreed, will be possibly addressed in the planned profile format refactoring?

perun/view/scatter_plot/factory.py
tfiedor: Great, note to myself we should fix/change this in bokeh_helpers (after discussion of the remodelling though). Probably would be nice to have two swatches: (1) e.g. viridian (for dataframes---number would be obtained by unique keys), (2) e.g. inferno (for lists---i.e. the number would be obtained by len()).

perun/view/scatter_plot/run.py
tfiedor: Yes, noted.

PR #34, opened 2018-02-28
Updated quadratic and constant model computations

JiriPavela:
Changes were made to the quadratic computation which now uses 3 coefficients and produces correct results (tested on sourced example). The constant model was re-classified as a derived model, i.e. its result (mainly r^2) is now computed based on a linear regression result on the given data. 

It is a heuristic - the computation may change in future to produce more precise data but tu current results seem to be close to what would be expected. E.g. linear line y = x produces constant r^2  = 0, constant line y = 3 produces constant r^2 = 1, data in between produce somewhat interpolated results.

### DIFF
--- a/perun/postprocess/regression_analysis/derived.py
+++ b/perun/postprocess/regression_analysis/derived.py
@@ -1,0 +1,100 @@
+"""Module with functions used for derived models computation.
+
+Some models may depend on results of regression models computed previously by
+standard regression analysis. Derived computations are more of a heuristics used
+for special cases.
+
+"""
+
+import perun.postprocess.regression_analysis.tools as tools
+
+
+def derived_const(analysis, const_ref, **_):
+    """The computation of a constant model based on a linear regression model.
+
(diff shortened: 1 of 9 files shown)


tfiedor:
Thanks for PR, will check ASAP the codeclimate changes and merge when everything is OK.

perun/postprocess/regression_analysis/derived.py
```diff
@@ -0,0 +1,100 @@
...
+        const['coeffs'] = [result['y_sum'] / result['pts_num'], 0]
+        const['uid'] = result['uid']
+        const['method'] = result['method']
+        yield const
+
+
+def _filter_by_models(analysis, models):
+    """Filters regression results by computed models.
+
+    Arguments:
```
tfiedor: Little note to future: I have switched to different notation of documenting arguments (i.e. the :param: stuff), because these are supported by sphinx, while the previous one has not. Eventually I will refactor everything in Perun, but for new function, please use the new notation instead :).

PR #37, opened 2018-03-15
SystemTap complexity collector

JiriPavela:
This pull request replaces the old gcc-compiler based complexity collector with the new SystemTap based collector. This will allow us to add tracing and probing points directly into the binary without the need of the source files. Note that this also adds new dependencies - most notably the SystemTap (stap) and -dbgsym version of kernel.


perun/collect/complexity/run.py
```diff
@@ -7,26 +7,25 @@
...
 import perun.logic.runner as runner
 import perun.utils.exceptions as exceptions
 
 # The profiling record template
-_ProfileRecord = collections.namedtuple('record', ['action', 'func', 'timestamp', 'size'])
+_ProfileRecord = collections.namedtuple('record', ['offset', 'func', 'timestamp'])
 
 # The collect phase status messages
 _COLLECTOR_STATUS_MSG = {
     0:  'OK',
```
tfiedor: Erm, can we change this to Enum? O:-)

perun/collect/complexity/run.py
```diff
@@ -39,174 +38,206 @@
...
+
+        print('Done.\n')
         return 0, _COLLECTOR_STATUS_MSG[0], dict(kwargs)
     # The "expected" exception types
     except (OSError, ValueError, subprocess.CalledProcessError,
-            UnicodeError, exceptions.UnexpectedPrototypeSyntaxError) as exception:
-        return 1, repr(exception), kwargs
+            UnicodeError, exceptions.StrategyNotImplemented) as exception:
+        print('Failed.\n')
+        return 1, str(exception), kwargs
```
tfiedor: Same as above, internally it would look better for you to use Enum (or maybe rather IntEnum), however I am afraid, that this will need to be changed in higher part of Perun. I will check it out and remodel the api, so some kind of enum can be returned.

perun/collect/complexity/run.py
```diff
@@ -39,174 +38,206 @@
...
-    return 0, _COLLECTOR_STATUS_MSG[0], dict(kwargs)
+    func_map = dict()
 
+    # Get the trace log path
+    try:
+        with open(kwargs['output'], 'r') as profile:
+
+            # Create demangled counterparts of the function names
+            demangler = shutil.which('c++filt')
+            if demangler is None:
```
tfiedor: demangler is None -> not demangler?

tfiedor: Also codacy does not like this, so would be good to call shlex.quote to guard against potential attack.

perun/collect/complexity/run.py
```diff
@@ -39,174 +38,206 @@
...
-        record(namedtuple): the _ProfileRecord tuple containing the record data
-        call_stack(list): the call stack with file records
-        resources(list): the list of resource dictionaries
-        address_map(dict): the function address: demangled name map
+    :param dict kwargs: complexity collector configuration parameters
+    :returns: tuple (int code value - nonzero for errors, path to the SystemTap output)
+    """
+    # Resolve the system tap path
+    stap = shutil.which('stap')
+    if stap is None:
```
tfiedor: Same as above "if not stap" seems better to me.

perun/collect/complexity/run.py
```diff
@@ -39,174 +38,206 @@
...
+    script_path, script_dir, _ = _get_path_dir_file(kwargs['script'])
+    # Create the output file and collection log
+    output = script_dir + kwargs['cmd_base'] + '_stap_record.txt'
+    with open(script_dir + kwargs['cmd_base'] + '_stap.log', 'w') as log:
+        # Start the collector
+        stap_runner = subprocess.Popen(('sudo', stap, '-v', script_path, '-o',
+                                        output, '-c', kwargs['cmd']),
+                                       cwd=script_dir, stderr=log, shell=False)
+        stap_runner.communicate()
+        if stap_runner.returncode != 0:
```
tfiedor: code = int(stap_runner.returncode != 0)?

perun/collect/complexity/strategy.py
```diff
@@ -0,0 +1,146 @@
...
+# The default global sampling for 'sample' strategies if global sampling is not set
+_DEFAULT_SAMPLE = 5
+
+
+def get_supported_strategies():
+    """Provides list of supported strategies.
+
+    :returns: list -- the names of supported strategies
+    """
+    return [key for key in _STRATEGIES.keys()]
```
tfiedor: Doesn't "list(_STRATEGIES.keys())" provide the same functionality and is clearer?

perun/collect/complexity/strategy.py
```diff
@@ -0,0 +1,146 @@
...
+                      where index is unique value representing the rule name
+
+    """
+
+    # Create samples from sampling list, filter <negative, 1> entries
+    samples = {samp['func']: (samp['sample'], idx) for idx, samp in
+               enumerate(sampling) if samp['sample'] > 1}
+
+    # Create samples for all remaining rules if needed
+    if global_sampling is not None and global_sampling > 1:
```
tfiedor: if global_sampling? Maybe instead?

PR #46, opened 2018-07-18
Trace collector architecture upgrade

JiriPavela:
This PR contains reworked architecture for the systemtap collector and static probes support. 

Specifically, the collector process is now being run as a background process, which allows to profile multiple binaries in one run (multi-process support will be added soon) and integrate e.g. automated testing scripts with perun performance profiling.

The static probes support allows the user to profile custom code segments / blocks that would not be normally profiled by simple function probes.


perun/collect/complexity/strategy.py
```diff
@@ -31,100 +35,139 @@ def get_default_strategy():
...
+    for probe in probes:
+        # Split the probe definition into pair or probes
+        delim = probe['name'].find('#')
+        if delim != -1:
+            probe['pair'] = probe['name'][delim + 1:]
+            probe['name'] = probe['name'][:delim]
+            result.append(probe)
+        elif probe['name'].endswith('_end') or probe['name'].endswith('_END'):
+            # Skip the end probes
+            continue
```
tfiedor: I think the if/elif/else flow can be simplified to just if/elif, by negating the elif condition and including the body of else (this will raise the coverage).

perun/collect/complexity/systemtap.py
```diff
@@ -0,0 +1,320 @@
...
+    output = os.path.join(kwargs['cmd_dir'], 'collect_record_{0}.txt'.format(kwargs['timestamp']))
+    log = os.path.join(kwargs['cmd_dir'], 'collect_log_{0}.txt'.format(kwargs['timestamp']))
+
+    with open(log, 'w') as logfile:
+        # Start the SystemTap process
+        print('Starting the SystemTap process... ', end='')
+        stap_runner, code = start_systemtap_in_background(script, output, logfile, **kwargs)
+        if code != Status.OK:
+            return code, None
+        print('Done')
```
tfiedor: Consider using log.done() from perun.utils.log

perun/collect/complexity/systemtap.py
```diff
@@ -0,0 +1,320 @@
...
+                # Take a break before the next status check
+                stap_process.wait(timeout=0.5)
+                # The process actually terminated which means that error occurred
+                return Status.STAP
+            except TimeoutExpired:
+                # Check process status and reload the log file
+                scanlog.seek(0)
+                # Read the last line of logfile and return if the systemtap is ready
+                last = ''
+                for line in scanlog:
```
tfiedor: last = '' if not len(line) else line[-1]? :)

perun/collect/complexity/systemtap.py
```diff
@@ -0,0 +1,320 @@
...
+def _parse_record(line):
+    """ Parses line from collector output into record tuple consisting of:
+        record type, call stack offset, rule name, timestamp and sequence.
+
+    :param str line: one line from the collection output
+    :returns namedtuple: the _TraceRecord tuple
+    """
+
+    # Split the line into = 'type' 'timestamp process' : 'offset' 'rule'
+    parts = line.partition(':')
```
tfiedor: Extracting the tuple returned by line.partition might be more readable, i.e. timestamp, _, right_section = line.partition(":"). Or something like that.

perun/utils/__init__.py
```diff
@@ -207,13 +207,14 @@ def run_safely_list_of_commands(cmd_list):
...
-def get_stdout_from_external_command(command):
+def get_stdout_from_external_command(command, **kwargs):
     """Runs external command with parameters, checks its output and provides its output.
 
     :param list command: list of arguments for command
+    :param kwargs: additional arguments for the subprocess invocation
     :return: string representation of output of command
     """
-    output = subprocess.check_output([c for c in command if c is not ''], stderr=subprocess.STDOUT)
+    output = subprocess.check_output([c for c in command if c is not ''], stderr=subprocess.STDOUT, **kwargs)
```
tfiedor: Is this safe? I have faith in you so I don't want to think in depth about this line of code, but Codacy reported it  (though I ignored it :P).

PR #53, opened 2019-02-06
ss/regressogram

xstupi00:
This pull request adds new postprocessor for non-parametric analysis. Implements first non-parametric method, concretely regressogram. Also adds tests for the new method and actualize documentation about the new postprocessor and regressogram method.


docs/postprocessors.rst
```diff
@@ -27,6 +27,11 @@ Perun's tool suite currently contains the following two postprocessors:
      which can be used for further postprocessing (e.g. by regression analysis) or to group
      similar amounts of resources.
 
+  4. :ref:`postprocessors-regressogram` (authored by **Simon Stupinsky**) same as above
+     mentioned *regression analysis*, attempts to find the fitting model for the dependent variable
+     based on another independent one. Regressogram represents the simplest method of non-parametric
```
tfiedor: Seems wrong: you are not finding the fitting model based on another independent one. Quite contrary we are trying to fit models through data only. Definitely think if this first sentence is true :-)

Btw. I think there should be comma before same

docs/postprocessors.rst
```diff
@@ -140,6 +145,57 @@ we can see that the best model for the amount of allocated memory depending on c
...
+
+.. _postprocessors-regressogram-examples:
+
+Examples
+""""""""
+
+    .. code-block:: json
+
+        {
+            "bin_stats": [
```
tfiedor: Not fond of: I think the "bin" is rather ambiguous description (clashes with "binary" used in many places). I understand you mean the window bin, but I'm wondering how we can change this to something different. After some thinking , how about using "bucket" instead? The same upplies to line 176

docs/postprocessors.rst
```diff
@@ -140,6 +145,57 @@ we can see that the best model for the amount of allocated memory depending on c
...
+            "uid": "linear::test2",
+            "bin_method": "doane",
+            "method": "regressogram",
+            "r_square": 0.7575757575757576,
+            "x_interval_end": 9.0,
+            "statistic": "mean",
+            "x_interval_start": 0.0
+        }
+
+The example above shows an example of profile post-processed by regressogram method (note
```
tfiedor: Fix: What is missing in this example is (1) explain very briefly what is the "doane" method (maybe with link) and (2) what are the bin stats.

perun/postprocess/regression_analysis/tools.py
```diff
@@ -169,3 +169,19 @@ def as_plot_y_dict(plot_y):
     :return: dictionary with key 'plot_y' set to plot_y
     """
     return dict(plot_y=plot_y)
+
+
+def add_analysis_to_profile(profile, analysis):
```
tfiedor: Suggestion: I approve of the refactor, however, I would rename the function to something more spot on, such as "add_model_to_profile" and "analysis" parameter to "model"

perun/postprocess/regressogram/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""
+Postprocessing of input profiles using the non-parametric method - regressogram.
```
tfiedor:
```suggestion
Postprocessing of input profiles using the non-parametric method: regressogram.
```

perun/postprocess/regressogram/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+This method serves for finding fitting models for trends in the captured
+profiling resources using by constant function at the individual parts of
```
tfiedor:
```suggestion
profiling resources using the constant function at the individual parts of
```

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
+"""
+Module with regressogram computational method and auxiliary methods at executing of this method.
+"""
+
+import numpy as np
+import numpy.lib.function_base as np_bin_selectors
+from click import BadParameter
```
tfiedor: Fix: don't use `from module import unit`. It is more error prone and less readable. I would probably even suggest to import numpy as "numpy" at line 5, but in these cases it is sometimes acceptable.

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+from scipy import stats
+from sklearn import metrics
+
+from perun.postprocess.regression_analysis.tools import validate_dictionary_keys
```
tfiedor: Fix: same as above: instead `import perun.postprocess.regression_analysis.tools as tools` and use qualified names.

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+    :param iter data_gen: the generator object with collected data (data provider generators)
+    :param dict configuration: the perun and option context
+    :return: list of dict: the computation results
+    """
+    # checking the presence of specific keys in individual methods
+    validate_dictionary_keys(configuration, _REQUIRED_KEYS, [])
+
+    # list of result of the analysis
+    analysis = []
+    for chunk in data_gen:
```
tfiedor: Fix: extract chunk[0], chunk[1], chunk[2] into appropriate variable names. If they are tuple you can use "for a, b, c in data_gen" and it should extract correctly.

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+    (independent variable) is covered by disjoint bins, and the value of a regressogram in
+    a bin is the mean/median of y-values (dependent variable) for the x-values inside that bin.
+
+    If the 'statistic' contains the int, then this number defines the number of bins. If the
+    'statistic' contains the string from the keys of <_BIN_SELECTORS>, then will be use the chosen
+    method to calculate the optimal bin width and consequently the number of bins.
+
+    :param list x_pts: the list of x points coordinates
+    :param list y_pts: the list of x points coordinates
+    :param str statistic: the statistic to compute
```
tfiedor: Suggestion: maybe "stat_function" or "statistic_function" would be more clear?

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+    :param str/int bins: the number of bins to calculate or the name of computational method
```
tfiedor: Suggestion: same as far above, I think it would be wiser to talk about e.g. "buckets" instead of "bins"

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+        'statistics': statistic,
+        'bin_stats': bin_stats.tolist(),
+        'x_interval_start': np.min(bin_edges),
+        'x_interval_end': np.max(bin_edges),
+        'y_interval_start': min(y_pts),
+        'r_square': metrics.r2_score(y_pts, [bin_stats[bin_number - 1] for bin_number in bin_numbers])
+    }
+
+
+def step(graph, x_pts, y_pts, graph_params):
```
tfiedor: Suggestion: rename to "render_step" or "render_step_function"

docs/postprocessors.rst
xstupi00: Yes, they were nonsense. I wrote it again, and I think it should be cool. Look at it, please.

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+    yy = list(y_pts) + list(y_pts)
+    yy[::2] = y_pts
+    yy[1::2] = y_pts
+    yy = yy[1:]
+    graph.line(xx, yy, color=graph_params.get('color'), line_width=graph_params.get('line_width'),
+               legend=graph_params.get('legend'))
+    return graph
+
+
+def choose_bin_sizes(ctx, param, value):
```
tfiedor: Suggestion: Maybe rename to "select_bin_size" (or "select_bucket_size")?

perun/postprocess/regressogram/methods.py
```diff
@@ -0,0 +1,135 @@
...
+    :param click.Context ctx: internal object that holds state relevant for the script execution at every single level
+    :param click.Option param: additive options from commands decorator
+    :param int/str value: the number of bins to calculate or the name of computational method
+    :raises click.BadParameter: in case the name of the method is unsupported
+    :return int: the number of bins selected by user or computed by given method
+    """
+    if value.isdigit():
+        return int(value)
+    elif value in _BIN_SELECTORS:
+        return value
```
tfiedor: Fix: I think this is error. Shouldn't this be `return _BIN_SELECTORS[value]() instead? And it should take some parameters according to the input profile.

The function should always return only one type and this one returns either int or string.

docs/postprocessors.rst
xstupi00: I have replaced all uses of 'bin/bins' for 'bucket/buckets', to make it unambiguous.

perun/postprocess/regressogram/run.py
```diff
@@ -0,0 +1,76 @@
+"""
+Postprocessor module with non-parametric analysis using the regressogram method.
+"""
+import click
+
+import perun.logic.runner as runner
+import perun.postprocess.regressogram.methods as methods
+from perun.postprocess.regression_analysis.data_provider import data_provider_mapper
```
tfiedor: Fix: again use absolute imports combined with as. `from` can be used e.g. for structures or constants, but not for functions.

perun/postprocess/regressogram/run.py
```diff
@@ -0,0 +1,76 @@
...
+    profile = add_analysis_to_profile(profile, analysis)
+
+    # Return the profile after the execution of regressogram method
+    return PostprocessStatus.OK, '', {'profile': profile}
+
+
+# TODO: The possibility of after postprocessing phase
+
+@click.command()
+@click.option('--bins', '-b', required=False, default=_DEFAULT_BINS_METHODS,
```
tfiedor: Suggestion: I'm thinking if it would be better to rename this to something longer, but I'm not sure at this moment to which. So lets keep this.

perun/postprocess/regressogram/run.py
```diff
@@ -0,0 +1,76 @@
...
+              multiple=False, callback=methods.choose_bin_sizes,
+              help=('Restricts the number of bins or method for its determination, to which '
+                    'will be placed the values of the selected statistics at regressogram method.'))
+@click.option('--statistic', '-s', type=click.Choice(['mean', 'median']),
```
tfiedor: Suggestion: Again this could be longer/more descriptive, but I have no viable suggestion at this moment.

perun/postprocess/regressogram/run.py
```diff
@@ -0,0 +1,76 @@
...
+              required=False, default=_DEFAULT_STATISTIC, multiple=False,
+              help='Will use the <statistic> to compute the values for points within each bin of regressogram.')
+@click.option('--depending-on', '-dp', 'per_key', default='structure-unit-size',
```
tfiedor: Fix: Remove? Where this is needed? I think only `--of` is required

perun/postprocess/regressogram/run.py
```diff
@@ -0,0 +1,76 @@
...
+
+        **Regressogram**: can be described such as step function (i.e. constant function
+        by parts). Regressogram uses the same basic idea as a histogram for density estimate.
+        This idea is in dividing the set of values of the independent variable (`<depending_on>`) into
+        intervals and the estimate of the point in concrete interval takes the mean/median of the
+        dependent variable (`<of_resource_key>`), respectively of its value on this sub-interval.
+        We currently use the `coefficient of determination` (:math:`R^2`) to measure the fitness of
+        regressogram. The models are stored as we can see an example below.
+
+    For more details about this approach of non-parametric analysis refer to :ref:`postprocessors-regressogram`.
```
tfiedor: Fix: I'm missing better description of `--bins` parameters, since there are different methods for computation of the buckets number. Please list them in this documentation, with very brief description).

perun/postprocess/regressogram/methods.py
```diff
@@ -1,17 +1,24 @@
...
+_REQUIRED_KEYS = ['bucket_method', 'statistic_function']
+
+
+def get_supported_methods():
+    """Provides all currently supported computational methods, to
+    estimate the optimal number of buckets, as a list of their names.
+
+    :returns list of str: the names of all supported methods
+    """
+    return [key for key in _BUCKET_SELECTORS.keys()]
```
tfiedor: Fix: change to `return list(_BUCKET_SELECTORS.keys())`

perun/postprocess/regressogram/methods.py
```diff
@@ -23,59 +30,64 @@ def compute(data_gen, configuration):
...
+    tools.validate_dictionary_keys(configuration, _REQUIRED_KEYS, [])
 
     # list of result of the analysis
     analysis = []
-    for chunk in data_gen:
-        result = regressogram(chunk[0], chunk[1], configuration['statistic'], configuration['bins'])
-        result['uid'] = chunk[2]
+    for x_pts, y_pts, uid in data_gen:
+        # Check whether the user gives as own number of buckets or select the method to its estimate
+        buckets = configuration['bucket_number'] if configuration.get('bucket_number') else configuration[
```
tfiedor: Suggestion: If the conditional is checking whether 'bucket_number' is in the configuration, then I would suggestion to simplify this to `buckets = configuration.get('bucket_number', configuration['bucket_method'])`, i.e. if 'bucket_number' is in configuration used it, otherwise as default use the computed one (though I guess it could call the underlying method)

xstupi00: It is a good idea, but _click_ configuration also includes the keys of parameters, which were not entered, with default value **None**. That's why the condition `configuration.get('bucket_number')` is valid due to the described behavior.

PR #54, opened 2019-02-12
WP: ss/moving average

xstupi00:
This pull request adds new postprocessor, which implemented the non-parametric method - moving average. Also, add tests for the new method.


docs/postprocessors.rst
```diff
@@ -31,9 +31,18 @@ Perun's tool suite currently contains the following four postprocessors:
      approach, is the simplest non-parametric estimator. This method trying to fit models through
      data by dividing the interval into N equal-width bucket and the resultant value in each bucket
      is equal to result of selected statistical aggregation function (mean/median) within the values
-     in the relevant bucket. In short, we can describe the regressogram as a render_step_function function
+     in the relevant bucket. In short, we can describe the regressogram as a step function
```
tfiedor: 🤣

perun/postprocess/moving_average/methods.py
```diff
@@ -0,0 +1,221 @@
...
+    """Provides all names of currently supported parameters to specify
+    the relationship to determine the smoothing parameter at Exponential
+    Moving Average.
+
+    :returns list of str: the names of all supported parameters
+    """
+    return list(_DECAY_PARAMS_INFO.keys())
+
+
+def window_width_change(window_width, r_square):
```
tfiedor: Fix: get_window_width_change or compute_window_width_change would be a little cleaner.

perun/postprocess/moving_average/methods.py
```diff
@@ -0,0 +1,221 @@
...
+
+    :param iter data_gen: the generator object with collected data (data provider generators)
+    :param dict configuration: the perun and option context
+    :return: list of dict: the computation results
+    """
+    # checking the presence of specific keys in individual methods
+    tools.validate_dictionary_keys(configuration, _METHOD_REQUIRED_KEYS[configuration['moving_method']], [])
+
+    # list of result of the analysis
+    analysis = []
```
tfiedor: Suggestion: I would consider renaming the "analysis" variable to "models" or something more appropriate.

perun/postprocess/moving_average/run.py
```diff
@@ -0,0 +1,223 @@
...
+                                                           configuration)
+
+    # Return the profile after the execution of moving average method
+    return PostprocessStatus.OK, '', {'profile': tools.add_models_to_profile(profile, moving_average_models)}
+
+
+# TODO: The possibility of after postprocessing phase
+
+
+def common_options(f):
```
tfiedor: Very strong suggestion: While this seems like good hack, I can't say I particularly like this solution. Consider moving this to the group instead and propagate through context.

The difference is in how the perun will be executed from cli:
1) perun postprocessby <postprocess_by_options> moving_average ema <common_opt> <ema_specific>
2) perun postprocessby <postprocess_by_options> moving_average <common_opt> ema <ema_specific>

I would say (2) is better and can be solved in Click (check other CLI functions, I'm sure I have this behaviour along with passing context solved already). Or maybe, screw this, finish the MA and I will refactor this myself (could be faster).

perun/postprocess/regression_analysis/tools.py
```diff
@@ -30,13 +30,9 @@ def validate_dictionary_keys(dictionary, required_keys, forbidden_keys):
     if not isinstance(dictionary, dict):
         raise exceptions.DictionaryKeysValidationFailed(dictionary, [], [])
     # Check all the required keys
-    for key in required_keys:
```
tfiedor: Suggestion: Nice hack, though I would keep the previous non-compressed loop. It is well readable, simple, no need to compress it :). Same suggestion applies to line 37

perun/view/scatter/factory.py
```diff
@@ -81,11 +82,12 @@ def generate_plot_data_slices(profile):
             yield uid_slice, []
 
 
-def draw_models(graph, models):
+def draw_models(graph, models, profile):
     """ Add models renderers to the graph.
 
     :param charts.Graph graph: the scatter plot without models
     :param list models: list of models to plot
+    :param profile: TODO
```
tfiedor: Fix: !!!

PR #56, opened 2019-03-05
ss/kernel regression

xstupi00:
This pull request adds new postprocessor for kernel regression. This postprocessor implements different approaches of this non-parametric regression: 

**kernel-regression:**
The first method of kernel regression is implemented with use python package _statsmodels_ and represents the clearly kernel-regression. This method allows the automatic selection of the bandwidth, using by _EstimatorSettings_. Offers to choose from **four** methods to automatic selection of bandwidth. For more info see Perun documentation, which is part of this PR also.

**kernel-smoothing:**
The second method implemented in this postprocessor is called _kernel-smoothing_. This method represents the kernel regression from _PyQt-Fit_ package, which allows another appropriately functionality. In this method can be selected from the **five types of kernel****, three regression methods** and **two methods** for automatic selection of bandwidth. More information are available in the Perun documentation.

**kernel-ridge:**
The last method, which are the part of this postprocessor implements the own class _KernelRidge_, which serves to executing this types of kernel regression. Exactly, this class implements the Nadaray-Watson kernel regression with automatic bandwidth selection via leave-one-out cross validation. This class is based on the class from the _sklearn.base_  modules. More info in the attached documentation.


perun/postprocess/kernel_regression/run.py
```diff
@@ -0,0 +1,555 @@
...
+        ctx.invoke(estimator_settings)
+
+
+# Supported modes at executing kernel regression:
+# - estimator-settings: with use EstimatorSettings and its arguments
+# - user-selection: bandwidth defined by user itself
+# - bandwidth_methods: bandwidth computed by helper method for its determine
+# - kernel-smoothing: provides the ability to choose a kernel and other methods
+# - kernel-ridge: set bandwidth by minimizing mean-squared error in leave-one-out cross validation from given range
+_SUPPORTED_MODES = [estimator_settings, user_selection, method_selection, kernel_smoothing, kernel_ridge]
```
tfiedor: Suggestion: I will leave the comment here. Since we have 5 subcommands, I would brainstorm about some different names (in particular for first three).

perun/view/scatter/factory.py
```diff
@@ -155,15 +155,24 @@ def create_moving_average_model(graph, model, profile, colour):
...
-    # Obtains the x-coordinates with the required uid to pair with current model
-    params = {'of_key': 'amount', 'per_key': model['per_key']}
-    for x_pts, _, uid in data_provider.data_provider_mapper(profile, **params):
-        if uid == model['uid']:
-            # Plot the model
-            graph.line(x=sorted(x_pts), y=model['bucket_stats'], line_color=colour, line_width=3.5, legend=legend)
-    return graph
+    def draw_model(y_pts):
+        # Obtains the x-coordinates with the required uid to pair with current model
+        params = {'of_key': 'amount', 'per_key': model['per_key']}
```
tfiedor: Suggestion: Consider extracting this to more rows, i.e. 
```
params = {
  ' of_key': ...
  'per_key':...
}
```

PR #60, opened 2019-03-29
Stats module

JiriPavela:
This PR will add new 'stats' module to the core of perun. This module allows us to store and manipulate various useful data or statistics within files in the .perun/stats directory. 

The .perun/stats directory is further divided into a hierarchy of directories based on the SHA-value of the vcs minor versions - thus, every minor version has its own 'stats' files. Each 'stats' file can contain multiple records which are distinguished by the supplied ID. The content of the files is also compressed to reduce the memory requirements.

### DIFF
--- a/perun/logic/stats.py
+++ b/perun/logic/stats.py
@@ -21,12 +21,13 @@ def get_stats_filename_as_source(profile, ignore_timestamp, minor_version=None):
     :param str profile: the profile identification, can be given as tag, sha value,
                         sha-path (path to tracked profile in obj) or source-path
     :param bool ignore_timestamp: if set to True, removes the 'date' component in the source name
-    :param str minor_version: representation of the minor version
+    :param str minor_version: representation of the minor version or None for HEAD
 
     :return str: the generated stats filename (not path!)
     """
     profile_index_entry = _find_profile_entry(profile, minor_version)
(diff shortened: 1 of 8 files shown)


perun/logic/stats.py
```diff
@@ -0,0 +1,344 @@
...
+
+    :param str minor_version: the minor version representation or None for HEAD
+
+    :return tuple: (bool representing the existence of directory, the directory path)
+    """
+    _, minor_dir = store.split_object_name(pcs.get_stats_directory(), minor_version)
+    return os.path.exists(minor_dir), minor_dir
+
+
+def _find_minor_index(minor_version):
```
tfiedor: Suggestion: Again this seems like it should be in index instead

perun/logic/stats.py
```diff
@@ -0,0 +1,344 @@
...
+    """
+    # Find the index file
+    _, index_file = store.split_object_name(pcs.get_object_directory(), minor_version)
+    if not os.path.exists(index_file):
+        raise exceptions.IndexNotFoundException(index_file)
+    return index_file
+
+
+@commands.lookup_minor_version
+def _find_profile_entry(profile, minor_version):
```
tfiedor: Suggestion: Again, candidate for being in index.

PR #61, opened 2019-04-23
Temporary module

JiriPavela:
This PR will add new internal perun storage at '.perun/tmp/' for various temporary data or files that are used during collection, postprocessing etc. The module is expected to be used mostly internally by the developers - however, the module also offers some CLI for other users, so that they can clean or list the contents in a comfortable way.

### DIFF
--- a/perun/logic/commands.py
+++ b/perun/logic/commands.py
@@ -896,6 +896,7 @@ def print_temp_files(root, **kwargs):
     """
     # Try to load the files in the root directory
     try:
+        temp.synchronize_index()
         tmp_files = temp.list_all_temps_with_details(root)
     except InvalidTempPathException as exc:
         print("Error: " + str(exc))
@@ -991,3 +992,10 @@ def delete_temps(path, ignore_protected, force, **kwargs):
     except (InvalidTempPathException, ProtectedTempException) as exc:
         # Invalid path or protected files encountered
         print("Error: " + str(exc))
+
+
+def sync_temps():
+    """Synchronizes the internal state of the index file so that it corresponds to some possible
+    manual changes in the directory by the user.
+    """
+    temp.synchronize_index()
--- a/perun/logic/commands.py
+++ b/perun/logic/commands.py
@@ -919,21 +919,22 @@ def print_temp_files(root, **kwargs):
     if not kwargs['no_total_size']:
         total_size = utils.format_file_size(sum(size for _, _, size in tmp_files))
         print('Total size of all temporary files: {}'.format(
-            termcolor.colored(total_size, TEXT_EMPH_COLOUR, attrs=TEXT_ATTRS))
+            _set_color(total_size, TEXT_EMPH_COLOUR, not kwargs['no_color']))
         )
 
     # Print the file records
     print_formatted_temp_files(tmp_files, not kwargs['no_file_size'],
-                               not kwargs['no_protection_level'])
+                               not kwargs['no_protection_level'], not kwargs['no_color'])
 
 
-def print_formatted_temp_files(records, show_size, show_protection):
+def print_formatted_temp_files(records, show_size, show_protection, use_color):
     """Format and print temporary file records as:
     size | protection level | path from tmp/ directory
 
     :param list records: the list of temporary file records as tuple (size, protection, path)
     :param bool show_size: flag indicating whether size for each file should be shown
     :param bool show_protection: if set to True, show the protection level of each file
+    :param bool use_color: if set to True, certain parts of the output will be colored
     """
     # Handle empty tmp/ dir
     if not records:
@@ -945,24 +946,39 @@ def print_formatted_temp_files(records, show_size, show_protection):
     for file_name, protection, size in records:
         # Print the size of each file
         if show_size:
-            print('{}'.format(termcolor.colored(utils.format_file_size(size), TEXT_EMPH_COLOUR)),
-                  end=termcolor.colored(' | ', TEXT_WARN_COLOUR))
+            print('{}'.format(_set_color(utils.format_file_size(size),
+                                         TEXT_EMPH_COLOUR, use_color)),
+                  end=_set_color(' | ', TEXT_WARN_COLOUR, use_color))
         # Print the protection level of each file
         if show_protection:
             if protection == temp.UNPROTECTED:
                 print('{}'.format(temp.UNPROTECTED),
-                      end=termcolor.colored(' | ', TEXT_WARN_COLOUR))
+                      end=_set_color(' | ', TEXT_WARN_COLOUR, use_color))
             else:
-                print('{}  '.format(termcolor.colored(temp.PROTECTED, TEXT_WARN_COLOUR)),
-                      end=termcolor.colored(' | ', TEXT_WARN_COLOUR))
+                print('{}  '.format(_set_color(temp.PROTECTED, TEXT_WARN_COLOUR, use_color)),
+                      end=_set_color(' | ', TEXT_WARN_COLOUR, use_color))
 
         # Print the file path, emphasize the directory to make it a bit more readable
         file_name = file_name[prefix:]
         file_dir = os.path.dirname(file_name)
         if file_dir:
             file_dir += os.sep
-            print('{}'.format(termcolor.colored(file_dir, TEXT_EMPH_COLOUR)), end='')
+            print('{}'.format(_set_color(file_dir, TEXT_EMPH_COLOUR, use_color)), end='')
         print('{}'.format(os.path.basename(file_name)))
+
+
+def _set_color(output, color, enable_coloring=True):
+    """Transforms the output to colored version.
+
+    :param str output: the output text that should be colored
+    :param str color: the color
+    :param bool enable_coloring: switch that allows to disable the coloring - the function is no-op
+
+    :return str: the new colored output (if enabled)
+    """
+    if enable_coloring:
+        return termcolor.colored(output, color, attrs=TEXT_ATTRS)
+    return output
 
 
 def delete_temps(path, ignore_protected, force, **kwargs):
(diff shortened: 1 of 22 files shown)


perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+    """Checks if the temporary file given by 'file_path' exists.
+
+    For details regarding the path format, see the module docstring.
+
+    :param str file_path: the temporary file
+
+    :return bool: True if the file exists
+    """
+    file_path = _join_to_tmp_path(file_path)
+    return os.path.exists(file_path) and os.path.isfile(file_path)
```
tfiedor: Note: While I see how this could be refactored (extract os.path._ functions to parameter), I am not sure if it is worth it...

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+
+
+def new_temp(file_path, content, json_format=False, protect=False, compress=False):
```
tfiedor: Suggestion: Strange name. Maybe "create_new_temporary"?

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+                content = store.read_and_deflate_chunk(tmp_handle)
+            else:
+                content = tmp_handle.read()
+            # Parse the json-formatted files
+            if json_format:
+                content = json.loads(content)
+        return content
+    # Handle possible errors
+    except (OSError, ValueError, zlib.error):
+        return None
```
tfiedor: Suggestion: Is it worth it to call log.warn() or log.error() here?

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+    """Clears the content and resets properties of the temporary file 'file_path'
+
+    For details regarding the path format, see the module docstring.
+
+    :param str file_path: the path to the temporary file
+    """
+    file_path = _join_to_tmp_path(file_path)
+    _is_tmp_file(file_path)
+    with open(file_path, 'w'):
+        pass
```
tfiedor: Suggestion: This pass seems suspicious? Is this pythonic way of clearing the file?

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+    For details regarding the path format, see the module docstring.
+
+    If the 'root' is not provided or None, the whole tmp/ folder is considered as the root.
+
+    :param str root: the path to the root folder
+
+    :return list: paths to all the files in the 'root' folder hierarchy
+    """
+    # Get the correct root
+    if root is None:
```
tfiedor: Suggestion: Maybe extract this to root = ... if ... else ... ?

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+        root = pcs.get_tmp_directory()
+    else:
+        root = _join_to_tmp_path(root)
+    _is_tmp_dir(root)
+    # Omit the index file
+    return [os.path.join(dirpath, file) for dirpath, _, files in os.walk(root) for file in files
```
tfiedor: Suggestion: reformat?
`return [`
`   os.path...`
`]`

perun/logic/commands.py
```diff
@@ -882,3 +886,132 @@ def load_profile_from_args(profile_name, minor_version):
...
+
+    :param str root: the path to the directory that should be listed
+    :param kwargs: additional parameters such as sorting, output formatting etc.
+    """
+    # Try to load the files in the root directory
+    try:
+        temp.synchronize_index()
+        tmp_files = temp.list_all_temps_with_details(root)
+    except InvalidTempPathException as exc:
+        print("Error: " + str(exc))
```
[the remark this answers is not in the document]
JiriPavela: Solved.

perun/logic/commands.py
```diff
@@ -882,3 +886,132 @@ def load_profile_from_args(profile_name, minor_version):
...
+        return
+
+    # Filter the files by protection level if it is set to show only certain group
+    if kwargs['filter_protection'] != 'all':
+        for name, level, size in list(tmp_files):
```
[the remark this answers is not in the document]
JiriPavela: The list() is needed so that I can traverse and modify (remove some elements) the tmp_list at the same time.

perun/logic/commands.py
```diff
@@ -882,3 +886,132 @@ def load_profile_from_args(profile_name, minor_version):
...
+        # Print the file path, emphasize the directory to make it a bit more readable
+        file_name = file_name[prefix:]
+        file_dir = os.path.dirname(file_name)
+        if file_dir:
+            file_dir += os.sep
+            print('{}'.format(_set_color(file_dir, TEXT_EMPH_COLOUR, use_color)), end='')
+        print('{}'.format(os.path.basename(file_name)))
+
+
+def _set_color(output, color, enable_coloring=True):
```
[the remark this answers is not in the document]
JiriPavela: Moved.

perun/logic/commands.py
```diff
@@ -882,3 +886,132 @@ def load_profile_from_args(profile_name, minor_version):
...
+            temp.delete_temp_file(path, ignore_protected, force)
+        elif temp.exists_temp_dir(path):
+            # We might delete only files or files + empty directories
+            if kwargs['keep_directories']:
+                temp.delete_all_temps(path, ignore_protected, force)
+            else:
+                temp.delete_temp_dir(path, ignore_protected, force)
+        # The supplied path does not exist, inform the user so they can correct the path
+        else:
+            print("Note: The supplied path '{}' does not exist, no files deleted"
```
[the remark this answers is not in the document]
JiriPavela: Solved.

perun/logic/commands.py
```diff
@@ -882,3 +886,132 @@ def load_profile_from_args(profile_name, minor_version):
...
+                  .format(temp.temp_path(path)))
+    except (InvalidTempPathException, ProtectedTempException) as exc:
+        # Invalid path or protected files encountered
+        print("Error: " + str(exc))
```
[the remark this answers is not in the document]
JiriPavela: Solved.

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+import os
+import json
+import zlib
+
+import perun.logic.pcs as pcs
+import perun.logic.store as store
+import perun.utils.exceptions as exceptions
+
+
+UNPROTECTED, PROTECTED = 'unprotected', 'protected'
```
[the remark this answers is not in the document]
JiriPavela: Solved.

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+        'reverse': False
+    },
+    'size': {
+        'pos': 2,
+        'reverse': True
+    }
+}
+
+
+def temp_path(path):
```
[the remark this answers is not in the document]
JiriPavela: Removed the additional layer.

perun/logic/temp.py
```diff
@@ -0,0 +1,653 @@
...
+    """Touches the given directory path.
+
+    For details regarding the path format, see the module docstring.
+
+    :param str dir_path: the directory path
+    """
+    # Append the path to the tmp/ directory
+    dir_path = _join_to_tmp_path(dir_path)
+    # Create the new directory(ies)
+    if not os.path.exists(dir_path):
```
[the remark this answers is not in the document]
JiriPavela: The store.touch_dir() works only for one layer of new directory (i.e. does not work for 'newdir/more_new/another_new') and the store.touch_dir_range() seems a bit needlessly complicated compared to simple os.makedirs. Maybe create new function store.touch_dirs() would be useful?

perun/logic/temp.py
JiriPavela: I agree, it would either create one more useless call layer in case of internal private function or complicate the use of the function for the users.

perun/logic/temp.py
JiriPavela: Solved.

perun/logic/temp.py
JiriPavela: Not really, reading temporary files is expected to be done only internally by other modules (the only operations usable through cli are listing and deletion of the files) and thus it seems a bit pointless to print internal failures to the user. Moreover, when such errors occur, there is pretty much only one option - scrap the file and overwrite it. Thus the simple `None` returned. However, it might be useful to log the exception into the log file without printing it to stdout/stderr.

perun/logic/temp.py
JiriPavela: There are some additional options that I've found such as `truncate()` for already open files or `open(path, mode).close()` for closed files which is this case. However, the usage of `with` seemed like the best option to me.

perun/logic/temp.py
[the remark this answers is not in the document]
tfiedor: Well, as I'm looking, the `store.touch_dir_range()` does something different, I would keep the function. But maybe extend the `touch_dir()` to `touch_dirs()` and work with some kind of generator? I'd say it is needless to have both `touch_dirs()` and `touch_dir()`, the first should be enough. That would solve everything.

perun/logic/temp.py
tfiedor: Ook, it seemed suspicious though

perun/logic/temp.py
JiriPavela: Solved.

perun/logic/temp.py
JiriPavela: Solved, however it still does not fit on one line :)

perun/logic/temp.py
tfiedor: Ook, you can leave it like this, I would probably use the `open().close()`, but this is probably more fail-safe.

perun/logic/temp.py
tfiedor: Erm, I meant it like this (should have format it like that from start ;) )
```suggestion
    return [
        os.path.join(dirpath, file) for dirpath, _, files in os.walk(root) for file in files
    ]
```

JiriPavela: Yes, that's exactly how I modified it, however there is one more condition in the second `for` so the list comprehension is still broken into two lines, but it looks a bit better :+1:

PR #64, opened 2019-07-23
Add new detection methods and changes related to them

xstupi00:
This pull request adds new detection methods: **Integral Comparison** and **Local Statistics**. This pull request also adds new detection strategies, which provide to perform the detection with different kinds of model: **best-model**, **all-models**, **best-param**, **best-nonparam**, **all-param**, **all-nonparam**. In this pull request was renamed the profile items: `x_interval_start` -> `x_start` and `x_interval_end` -> `x_end`. Further, this pull request contains the reformatting of code style according to `Pylint` issues in non-parametric post-processors. In the end, this pull request contains many other minor changes related to the all mentioned tasks.


perun/postprocess/moving_average/run.py
```diff
@@ -127,41 +139,53 @@ def simple_moving_median(ctx, **kwargs):
...
+        observations faster, the small value to the contrary. Finally, to calculate the current
+        value of EMA is used the relevant formula. It is important do not confuse **Exponential**
+        Moving Average with **Simple** Moving Average. An **Exponential** Moving Average behaves
+        quite differently from the second mentioned method, because it is the function of weighting
+        factor or length of the average.
     """
-    kwargs.update({'moving_method': 'ema', 'window_width': kwargs['decay'][1], 'decay': kwargs['decay'][0]})
+    kwargs.update({
+        'moving_method': 'ema',
+        'window_width': kwargs['decay'][1],
```
tfiedor: Suggestion: This may be used on several places, so maybe it would be worthy to extract to namedtuple, like `Decay`

perun/postprocess/moving_average/run.py
```diff
@@ -127,41 +139,53 @@ def simple_moving_median(ctx, **kwargs):
...
     kwargs.update(ctx.parent.params)
     runner.run_postprocessor_on_profile(ctx.obj, 'moving_average', kwargs)
 
 
 @click.group(invoke_without_command=True)
 @click.option('--min_periods', '-mp', type=click.IntRange(min=1, max=None),
-              help='Provides the minimum number of observations in window required to have a value. '
-                   'If the number of possible observations smaller then result is NaN.')
+              help='Provides the minimum number of observations in window required to have a value.'
+                   ' If the number of possible observations smaller then result is NaN.')
```
tfiedor:
```suggestion
                   ' If the number of possible observations is smaller then the result is set to NaN.')
```

perun/postprocess/regression_analysis/methods.py
```diff
@@ -11,7 +11,7 @@
 import perun.postprocess.regression_analysis.tools as tools
 
 
-def get_supported_methods():
+def get_param_methods():
```
tfiedor: Suggestion: `get_supported_parametric_methods` or `get_supported_param_methods`. I assume this was renamed to avoid clash with non-param methods. However, I'm missing the supported keyword.

perun/postprocess/regressogram/methods.py
```diff
@@ -12,7 +12,16 @@
 _REQUIRED_KEYS = ['bucket_method', 'statistic_function']
 
 
-def get_supported_methods():
+def get_nparam_methods():
```
tfiedor: Suggestion: I see (continuation of previous comment), I would rename this to `get_supported_nparam_methods` ;)

perun/profile/factory.py
```diff
@@ -72,10 +77,11 @@ def update_resources(self, resource_list, resource_type='list', clear_existing_r
         elif resource_type == 'snapshots':
             # Resources are in type of [{'time': _, 'resources': []}
             for i, snapshot in enumerate(resource_list):
-                self._translate_resources(snapshot['resources'], {
-                    'snapshot': i,
-                    'time': snapshot.get('time', '0.0')}
-                )
+                self._translate_resources(snapshot['resources'],
```
tfiedor: Fix: This indentation seems weird. Either keep the previous one or extract to variable.

perun/profile/factory.py
```diff
@@ -203,7 +211,26 @@ def all_resources(self):
                 snapshot_number = collectable_properties.get('snapshot', 0)
                 yield snapshot_number, collectable_properties
 
-    def all_models(self):
+    def get_models(self, models_strategy):
```
tfiedor: Fix: The name is too similar to `all_models`, I would be confused which one to use. Rename this e.g. to `all_filtered_models` or `filtered_all_models`.

perun/utils/log.py
```diff
@@ -411,33 +484,55 @@ def keygetter(item):
...
             cprint("$ {}".format(cmd), CHANGE_CMD_COLOUR, attrs=['bold'])
             print(')')
+
+            # Print information about the change on the partial intervals (only at Local-Statistics)
+            if deg_info.partial_intervals is not None:
+                print_partial_intervals()
     print("")
 
 
+def aggregate_intervals(intervals):
```
tfiedor: Fix: Missing docstring.

perun/utils/log.py
```diff
@@ -401,6 +404,76 @@ def keygetter(item):
         """
         return item[0].location
 
+    def print_models_info():
```
tfiedor: Suggestion: Too much nested functions...

perun/check/local_statistics.py
```diff
@@ -0,0 +1,355 @@
...
+    :return tuple: (change states on individual sub-intervals, average relative error)
+    """
+    # create vectorized functions which take a np.arrays as inputs and perform actions over it
+    compare_diffs = np.vectorize(compare_diff_values)
+    classify_change = np.vectorize(check_helpers.classify_change)
+    # initialize np.arrays about the size of sub-intervals from given stats
+    change_score = np.full_like(base_stats.get(list(base_stats.keys())[0]), 0)
+    rel_error = np.full_like(base_stats.get(list(targ_stats.keys())[0]), 0)
+    # iteration over each metric in the given stats
+    for base_stat in base_stats.items():
```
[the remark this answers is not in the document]
xstupi00: Because `baseline_stats` and `target_stats` are dictionaries and I want to obtain the items with the same keys from both dictionaries. Therefore I iterate over the `baseline_stats` and according to the key of the current item I get the relevant item from the `target_stats`.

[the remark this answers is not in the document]
tfiedor: Aw yeah, my bad, I made some hasty assumptions :). Then I propose to extract this to `for (base_stat_key, base_stat_value)` it should not make anything too long and will be better readable.

perun/check/factory.py
```diff
@@ -279,3 +314,78 @@ def get_strategies_for(profile):
             method = parse_strategy(strategy['method'])
             already_applied_strategies.append(method)
             yield method
+
+
+def run_detection_for_profiles(detection_method, base_profile, targ_profile, models_strategy):
```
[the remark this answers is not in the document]
xstupi00: It is the confusing name of the function. This function is called by detection methods, which works with all model kinds, i.e. methods **Integral Comparison** and **Local Statistics**, which are called by function **degradation_in_profiles**. 

So the sequence of calls is the following: `degradation_in_profiles` -> `integral_comparison/local_statistics` -> `run_detection_with_strategy` (**new name**) -> `run_detection_for_obtained_models` (**new name**) -> `execute_analysis` at specific detection method.

Is this all right now?

perun/check/local_statistics.py
```diff
@@ -0,0 +1,355 @@
...
+    x_pts, base_y_pts, targ_y_pts = preprocess_models(
+        base_model, kwargs.get('targ_profile'), targ_model
+    )
+
+    base_stats, _ = compute_window_stats(x_pts, base_y_pts)
+    targ_stats, x_pts = compute_window_stats(x_pts, targ_y_pts)
+    change_info, partial_rel_error = classify_stats_diff(base_stats, targ_stats)
+
+    x_pts = np.append(x_pts, [x_pts[0]], axis=1) if x_pts.size == 1 else x_pts
+    x_pts_even = x_pts[:, 0::2].reshape(-1, x_pts.size // 2)[0].round(2)
```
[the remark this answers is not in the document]
xstupi00: Yes, it can be extracted to the function, but currently, it will be called only in thee two lines of code. So, I think, that not worth it.

perun/check/factory.py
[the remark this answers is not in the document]
tfiedor: Well, it is better, let me think a little. Rename the second one to `run_detection_for_models`. I'm also thinking that this could be private (i.e. with `_` prefix) so it is clear these are mostly meant as "helper functions" and not be directly called. However, it could lead to warnings in IDE/lint/etc.

perun/check/local_statistics.py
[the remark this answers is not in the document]
tfiedor: Ook, leave it then.

perun/postprocess/moving_average/run.py
xstupi00: I think that is no needed because there is single direct access to items of these parameters and further in the code are used only items of the dictionary. Moreover, the item `window_width` is uniform at all moving-average methods and this change would probably run the wave of the other changes.

tfiedor: Erm, I'm talking only about `kwargs['decay'], which currently contains tuple, and it could be instead namedtuple, so you would write `kwargs['decay'].rate` or something like that, I don't know what is in `[0]` and in `[1]`. Just consider if `kwargs['decay'][X]` is used in more places.

perun/check/factory.py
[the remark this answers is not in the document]
xstupi00: I think the IDE/lint and som other static checkers, will not be able to find this issue because the function is out of class. Therefore I try to add the `_` prefix.

perun/postprocess/moving_average/run.py
xstupi00: Change to the `NamedTuple` would mislead the `click`, because now is used the type of this argument `click.Tuple` and moreover `kwargs['decay]` is used only in this place and further are used only `window_width` and `decay` items.

tfiedor: AFAIK you can use in click custom classes as well, but if it is used only in this place, then it would lead to more changes and problem. So keep it like this then.

PR #70, opened 2020-02-11
ss/fixes

xstupi00:
This PR adds different fixes, which were occurred at testing the whole framework. Between the main change belongs to the preprocessing of non-parametric models at `integral-comparison` detection method. Further, this PR adds automatically removing of files created at testing and corrects some other minor issues, which were discovered occasionally. The last change, which is added in this PR., is the option enter the own name of the file, that will be created at the output of the collecting phase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
-    - "3.4"
+    - "3.5"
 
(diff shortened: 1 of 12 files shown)


tfiedor:
The build fails in travis.

PR #73, opened 2020-09-01
Callgraph-based approach of coverage analysis in fuzz testing.

xlisci02:
This pull request adds a new approach to the existing solution of coverage analysis within performance fuzzing: static call graph analysis. Using new `gcov` JSON output format we can associate the coverage data with the call graph nodes easily and effectively and therefore we obtain more sophisticated coverage-performance indicator. This approach should improve the ability of recognition and finding new workloads which badly affect the SUT.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,9 @@ install:
     - make install
 
 before_script:
-    - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.9 90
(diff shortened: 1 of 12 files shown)


fche:
BTW, see also systemtap's "stap-profile-annotate" gadget.

https://sourceware.org/git/?p=systemtap.git;a=blob;f=stap-profile-annotate.in;hb=HEAD

PR #90, opened 2022-12-29
profile/factory: Import MutableMapping base class from collections.abc

HarryMichal:
Importing abstract base classes from collections has been marked as
deprecated in Python 3.3[0] in favor of importing from collections.abc.
The deprecation has been finished in Python 3.9[1]

[0] https://github.com/python/cpython/issues/55294 [1] https://github.com/python/cpython/issues/70176


JiriPavela:
Good catch, I fixed this issue as part of fixing support for new Python versions in #101. Many thanks for finding the issue tho!

Issue #95, opened 2023-03-07
Re-enable and update automation via GitHub Actions


HarryMichal:
Nobody wants to do manual testing and deployments. Use the system we have available. :)

Using Tox (configured with all versions of Python we want to test against) for testing could simplify the action and help with coverage.

JiriPavela:
Fully completed by #113.

Issue #92, opened 2023-03-07
Move from setup.py to pyproject.toml for setuptools


HarryMichal:
### Description of the feature

<TODO>

### Perun version

`0.20.4`

JiriPavela:
Thanks for bringing this up, fixed in #98.

PR #98, opened 2023-08-14
Migration to pyproject.toml and tox

JiriPavela:
This PR removes the obsolete setup.py project configuration and substitutes it with the new pyproject.toml configuration file which is specified by multiple PEPs. The pyproject.toml also contains tool-specific configuration (e.g., for pylint, coverage, ...).

Moreover, a tox configuration has been created and tested so that we can easily automate testing, linting and type checking both locally and in GH Actions. Currently, the only python version supported by tox is 3.8 due to complicated dependency conflicts in higher python versions.

This PR solves #92 and partially #95 (GH Actions are not updated to use Tox yet).


perun/collect/memory/malloc.c
```diff
@@ -168,7 +168,8 @@ __attribute__ ((constructor)) void initialize (void) {
 __attribute__((destructor)) void finalize (void) {
     if(logFile != NULL){
         fprintf(logFile, "EXIT %fs\n", clock() / (double)CLOCKS_PER_SEC);
-        fclose(logFile);
+        // This is causing segfaults for some reason, hotfix
```
tfiedor:
```suggestion
        // FIXME: This is causing segfaults for some reason, hotfix
```
This way it won't go fully ingored.

pyproject.toml
```diff
@@ -0,0 +1,112 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "perun"
+description = "Perun: Lightweight Performance Version System"
+version = "0.20.4"
```
tfiedor, line 8: Is it possible to extract the version somehow? It's is not ideal for it to be as hardcoded.

tfiedor (approved):
I think this is fine, thank you for solving this issue.

However, I would ask you one more change: please change README according to these changes, since now building, running of tests, etc. has maybe changed.

perun/collect/memory/malloc.c
JiriPavela: Fixed.

JiriPavela: I have updated the README instructions a bit. Most of the instructions remained as they were, as the installation is usually done through the Makefile, which was updated accordingly. However, I have removed the outdated `perun init` command in both install/dev, and included some basic Tox instructions.

pyproject.toml
JiriPavela: Good catch, seems like I forgot to de-duplicate the versions before doing the PR. So the `pyproject.toml` file is actually the preferred place to hard-code the package version, as far as I understand. Project/package metadata should be stored in the toml file to make them available for package distribution software. So instead, I have made the `__init__.py` `__version__` attribute dynamic such that it reads the version from the installed package (see [this SO answer](https://stackoverflow.com/a/72168209)).

README.md
```diff
@@ -191,6 +189,18 @@ by the installation.
...
+Python versions, as well as run static type checking, code style linting and generating
+documentation:
+
+    tox run
+
+If you wish to test only against a single Python version, run Tox as:
+
+    tox run -e pyXY
+
+where `XY` is the python version you wish to test.
```
tfiedor: Suggestion: write this as `py3.X` maybe? and mention which version we currently safely support?

tfiedor (approved):

README.md
JiriPavela: The `pyXY` is actually a naming convention used by Tox to automatically deduce the Python interpreter version (see https://tox.wiki/en/latest/config.html#basepython), but it seems that it understands `pyX.Y` as well (no mention of this in docs tho?), so I updated it. As for the supported versions, I have updated the README with the command to show available Tox environments.

tfiedor: OK, that's fine then.

Issue #100, opened 2023-08-14
Custom bkcharts dependency conflict in new Python versions


JiriPavela:
Our custom bkcharts package wheel is causing dependency conflicts when trying to install Perun with Python3.9+. The bkcharts wheel is not compatible with Pandas >= 1.0.0 and this results in an unsuccessful attempt to build Pandas wheel for Python3.9.

[Bkcharts](https://github.com/bokeh/bkcharts) is no longer maintained and [HoloViews](https://github.com/holoviz/holoviews) is the successor of bkcharts. We should check if our original bkcharts plots can be transitioned to holoviews easily. 

This issue must be resolved in order to resolve #94 .

JiriPavela:
Fixed in #101.

Issue #99, opened 2023-08-14
Custom PyQt_Fit dependency conflict in new Python versions.


JiriPavela:
Our custom PyQt_Fit package wheel is causing dependency conflicts when trying to install Perun with Python3.9+. PyQt_Fit requires old versions of numpy and Cython, which in turn block new versions of Pandas (<1.0.0) that are not available as wheels in 3.9+.

PyQt_Fit is used only in [perun/postprocess/kernel_regression/methods.py](https://github.com/tfiedor/perun/blob/51593d9c2fd37b70ae8ba959e880d4a0db804204/perun/postprocess/kernel_regression/methods.py) and transitioning to non-parametric kernel regressions in other public packages might not be as straightforward. A possible hotfix would be to copy the pyqt_fit code directly into Perun (with a link to the original repo and author).

This issue must be resolved in order to resolve #94 .

JiriPavela:
Fixed in #101.

PR #101, opened 2023-08-23
Fix dependency hell and add support for newer Python versions

### LINKED ISSUE #94: Enable the use of modern versions of Python
I believe that any version above 3.8 is untested and may have issues with either core code or some of our dependencies. Playing with dependencies will be most likely needed.

JiriPavela:
This PR fixes Perun's dependency hell and, as a consequence, unblocks support for newer Python versions.

- [PyQt-Fit ](https://github.com/PierreBdR/PyQt-fit) custom wheel has been replaced with a local port by extracting the relevant source code (the original authorship has been mentioned in the module docstring). Transition to an actively developed package would be nice but is much more difficult.
- [Bkcharts](https://github.com/bokeh/bkcharts) custom wheel has been replaced with [Holoviews](https://github.com/holoviz/holoviews) due to bkcharts being no longer maintained.

Those two custom wheels have been blocking newer numpy and pandas versions, hence preventing us from supporting newer Python versions.

- All dependencies' constraints have been bumped to the newest possible versions that still support Python 3.8.
  - Obsolete and no longer used/needed dependencies have been removed.
- Python versions 3.8, 3.9, 3.10 and 3.11 have been tested and failing tests have been fixed.

This PR fixes #94, #99, #100 and supersedes PRs #89, #90.


perun/logic/commands.py
```diff
@@ -11,8 +11,6 @@
 
 from operator import itemgetter
 
-import colorama
```
tfiedor, line 14: What is the issue with colorama? Why was it removed?

perun/postprocess/regressogram/methods.py
```diff
@@ -113,29 +113,6 @@ def regressogram(x_pts, y_pts, statistic_function, buckets):
         )
     }
 
-
-def render_step_function(graph, x_pts, y_pts, graph_params):
```
tfiedor, line 117: What is the issue with this function?

perun/thirdparty/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Package for custom ports of third-party packages, libraries or modules.
+
+Some packages, libraries or modules have become unmaintained over the time. This package aims to
+centralize those dependencies (perhaps modified or fixed to be compatible with new Python versions)
+so that we can continue using them.
```
tfiedor, line 5: Can you please add some short thank you to all authors, whose code we are including in Perun? "We thank all the authors for their contribution to Python community and for allowing us to use them in our Perun project"

perun/utils/view_helpers.py
```diff
@@ -0,0 +1,206 @@
+"""Collection of helper functions for working with bokeh graphs"""
+from __future__ import annotations
```
tfiedor, line 2: Can you elaborate, why you require this import?

perun/utils/view_helpers.py
```diff
@@ -0,0 +1,206 @@
...
+
+import perun.profile.helpers as profiles
+from perun.utils import log
+
+if TYPE_CHECKING:
+    import pandas as pd
+    from perun.profile.factory import Profile
+
+
+__author__ = "Tomas Fiedor"
```
tfiedor, line 25: Ehm.

perun/utils/view_helpers.py
```diff
@@ -0,0 +1,206 @@
...
+    :param profile: a Perun profile that will be processed.
+    :param filename: output filename for the graph.
+    :param view_in_browser: ``True`` if the created graph should be displayed in registered browser
+           after it is constructed.
+    :param func: function used for aggregation of the data.
+    :param of_key: key that will be aggregated in the graph.
+    :param kwargs: rest of the keyword arguments.
+    :raises AttributeError: when the factory_module is missing some of the functions.
+    """
+    profiles.is_key_aggregatable_by(profile, func, of_key, "of_key")
```
tfiedor, line 205: Should this be in some `if` or something? Or does it fire exception? Then it should be caught maybe with some error?

tfiedor (approved):
Awesome, thank you very much for solving this, we are one step closer to bringing Perun back.

perun/logic/commands.py
JiriPavela: I removed colorama during the dependencies pruning as it felt a bit unnecessary in the current state of Perun. Since it is currently not possible to deploy or test Perun on Windows machines (and it will likely require a significant amount of work to do so), having a package that is aimed at fixing windows terminal output seems like needless dependency. Especially so since we are not able to test if it works as intended or not.

perun/postprocess/regressogram/methods.py
JiriPavela: Nothing really, the function was just moved to the scatter/factory.py module as it contains code related to plotting and not computing the postprocessing model itself.

perun/logic/commands.py
tfiedor: Ook, fine decision.

perun/postprocess/regressogram/methods.py
tfiedor: Found her later, fine decision.

perun/utils/view_helpers.py
JiriPavela: Especially in older Python versions, this import is needed to enable certain typing syntax sugar (e.g., using | instead of Union, or direct references of types in type aliases and type variables, ...). In this particular file, however, this import is needed so that pylint does not report false alarms regarding typing syntax for 3.8.

tfiedor: That's neat, I will use it as well, since I wanted to use `|` as well.

Maybe consider adding comment `# this imports new typing annotations such as | operator for Union of types` or something like this.

perun/utils/view_helpers.py
JiriPavela: Although Git shows this code as my addition, the file was originally called bokeh_helpers.py and this line was authored by you :) . I just updated and renamed the file to better suit the Holoviews needs. On a side note, I feel like the __author__ tags on a per-file basis are becoming a bit obsolete and unnecessary maybe? Maintaining those "tags" requires manual effort and is generally error-prone (a lot of modules were updated or changed by multiple people over time). Why not simply rely on git blame to deduce authorship on a line by line basis?

tfiedor: Yes, I am thinking the same, I think the issue is that it was automatically added at the begining, I kept it, and rest of you simply copied this :D.

perun/utils/view_helpers.py
JiriPavela: The function raises an InvalidParameterException that has a suitable user-friendly string representation and has been used to report click parameter errors to the users so far. Catching the error would not make much sense as it would be re-raised anyway. Note that this was taken as-is from the original boken_helpers.py.

tfiedor: Yeah, I get it some parts are taken from previous code, and might be caused by me :D. I wonder if it is worth either (1) adding comment `# this raises exception` or (2) renaming the function? The latter would require some more digging though how it is used (probably rather in `if`s). What do you think?

perun/thirdparty/__init__.py
JiriPavela: Fixed.

perun/utils/view_helpers.py
JiriPavela: Although I can see the usefulness of such comment, it would quickly become tedious to maintain, I think. In my experience with typed codebases, this import is very common when using some of the more recent typing features and the comments would quickly fall out of date when bumping the minimal supported Python version of Perun.

perun/utils/view_helpers.py
JiriPavela: In that case, I'd say let's document it in the docstring alongside the `AttributeError` (added bonus is that Sphinx supports it and it will appear in the docs). Although the `is_key_aggregatable_by` method has it documented, this method does not mention it.

perun/utils/view_helpers.py
tfiedor: Yes, I wasn't so sure with it as well. I also consider some extremely short comment on the precise line, such as `from __future__ import annotation # import typing for backward compatibility`, but I guess it will suffice. I myself now know about this trick, and future contributors will surely learn it as well. Resolved.

perun/utils/view_helpers.py
tfiedor: Ook, I consider everything fully discussed. Feel free to merge it yourself, once you fix what we have aggreed upon.

JiriPavela: I updated the method docstring to mention the exception and its cause. Would that work for you?

tfiedor: Yes

PR #105, opened 2023-08-27
Removal of authorship tags in source files.

### LINKED ISSUE #104: Remove obsolete comments (authorship)
The code is blighted with nasty old obsolete comments and stuff, such as `__authors__`. We should remove it and never introduce it again.

JiriPavela:
This PR removes authorship and co-authorship "tags" in source files as they are out of date and tedious to maintain. Let's rely on git blame from now on.

Fixes #104.

### DIFF
--- a/perun/templates/postprocess.__init__.jinja2
+++ b/perun/templates/postprocess.__init__.jinja2
@@ -4,7 +4,3 @@ TODO: Write long docstring of the new postprocessor
 """
 
 SUPPORTED_PROFILES = ["{{ supported_types|join('\", \"') }}"]
-
-{% if author is defined %}
-__author__ = '{{ author }}'
-{% endif %}
--- a/docs/_static/templates/collectors_init.py
+++ b/docs/_static/templates/collectors_init.py
@@ -5,4 +5,3 @@ COLLECTOR_DEFAULT_UNITS = {
     'type': 'unit'
 }
 
-__author__ = 'You!'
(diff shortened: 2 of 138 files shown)


perun/templates/postprocess.__init__.jinja2
```diff
@@ -4,7 +4,3 @@ TODO: Write long docstring of the new postprocessor
 """
 
 SUPPORTED_PROFILES = ["{{ supported_types|join('\", \"') }}"]
-
-{% if author is defined %}
```
tfiedor, line 8: This is actually bound to a certain parameter in CLI in (I believe) utils_group? Can you remove the parameter as well in this PR? Maybe you did, but there is so much changes, that I just skimmed it.

tfiedor (approved):
That was fast. There is just one small issues I believe: removing __author__ from template is bound to some CLI parameter, which is hence useless and should be removed as well. Thanks.

perun/templates/postprocess.__init__.jinja2
JiriPavela: Oh, good catch. I think I found it.

JiriPavela: The new commit should contain a fix.

tfiedor: Good, feel free to merge.

PR #114, opened 2023-09-24
tests: Use init.defaultBranch over 'master' as the default branch name

martymichal:
In git 2.28 it became possible to set the default branch name[0] and having it set to a different value than 'master' can cause issues. Get the default value from the config and if not set, fall back to 'master'.

[0] https://github.com/git/git/blob/master/Documentation/RelNotes/2.28.0.txt
[1] https://sfconservancy.org/news/2020/jun/23/gitbranchname

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -248,6 +248,9 @@ def full_profiles():
 def pcs_with_degradations():
     """
     """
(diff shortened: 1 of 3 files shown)


JiriPavela (approved):
Good catch, all tests seem to work properly so I'm approving the merge.

PR #115, opened 2023-09-24
Add typing annotation to Perun

tfiedor:
This PR:
  1. Fixes minor issues, and refactor selected parts of codes;
  2. Adds proper typing to all (but few) modules wrt to newest versions of the Python.


tfiedor:
There is still missing typing of `perun/trace/*` and `perun/thirdparty`. @JiriPavela do you want to fix them yourself?

perun/check/factory.py
```diff
@@ -289,10 +314,10 @@ def parse_strategy(strategy):
     return short_strings.get(strategy, strategy)
 
 
-def get_strategies_for(profile):
+def get_strategies_for(profile: Profile) -> Iterable[str]:
     """Retrieves the best strategy for the given profile configuration
 
-    :param ProfileInfo profile: Profile information with configuration tuple
+    :param Profileprofile: Profile information with configuration tuple
```
JiriPavela: Seems like a small typo here. However, this leads me to a more general issue: Should we still continue using the old type hints in docstrings now that our codebase is mostly type annotated? This seems a bit redundant to me.

tfiedor: I kept them there, since it seems that some checkers/IDEs/etc might use them (I think PyCharm---though it might be older version---complained with highlight and it annoyed me. But we might stop use them and slowly remove them.

perun/check/integral_comparison.py
```diff
@@ -19,7 +26,7 @@
 _INTEGRATE_DIFF_CHANGE = .25
 
 
-def compute_param_integral(model):
+def compute_param_integral(model: perun.utils.structs.ModelRecord) -> float:
```
JiriPavela: Is there a reason why the ModelRecord uses a fully specified path? Was there perhaps a problem with using just `ModelRecord`?

perun/check/linear_regression.py
```diff
@@ -27,9 +40,14 @@ def linear_regression(baseline_profile, target_profile, **_):
...
 def exec_linear_regression(
-        uid, baseline_x_pts, lin_abs_error, threshold, linear_diff_b1,
-        baseline_model, target_model, baseline_profile
-):
+        uid: str,
+        baseline_x_pts: npt.NDArray[numpy.float64] | list[float],
+        lin_abs_error: npt.NDArray[numpy.float64] | list[float],
+        threshold: int,
+        linear_diff_b1: float,
+        baseline_model: perun.utils.structs.ModelRecord, target_model: perun.utils.structs.ModelRecord,
```
JiriPavela: Another occurrence of the same thing as in `integral_comparison.py`.

perun/postprocess/normalizer/run.py
```diff
@@ -30,7 +33,7 @@ def normalize_resources(resources):
     :param list resources: list of resources
     """
     # First compute maximas per each type
-    maximum_per_type = {}
+    maximum_per_type = {}  # type: dict[str, int]
```
JiriPavela: Why not annotate this as `maximum_per_type: dict[str, int] = {}`. Was there some issue with this?

perun/postprocess/regression_analysis/data_provider.py
```diff
@@ -40,8 +47,8 @@ def generic_profile_provider(profile, of_key, per_key, **_):
 
     # Sort the dictionaries by function name for easier traversing
     resources = sorted(resources, key=resource_sort_key)
-    x_points_list = []
-    y_points_list = []
+    x_points_list = []  # type: list[float]
```
JiriPavela: The same thing as in normalizer.

perun/utils/structs.py
```diff
@@ -300,7 +329,7 @@ def __init__(self, *args):
         ordered = len(self.__class__.__members__) + 1
         self.order = ordered
 
-    def __ge__(self, other):
+    def __ge__(self, other: Any) -> bool:
```
JiriPavela: Just a minor remark. I think all the comparison methods are usually typed with `other` being `object` rather than `Any` in the typeshed.

pyproject.toml
```diff
@@ -163,4 +196,20 @@ omit = [
...
+]
+
+[tool.coverage.report]
+exclude_also = [
+    # Exclude type checking specific lines, that are never executed
+    "if TYPE_CHECKING:",
+
+    # Exclude protocols, that are used for type checking
+    "class CallableDetectionMethod(Protocol):",
+    "class Comparable(Protocol):",
```
JiriPavela, line 208: This protocol exclusion list seems like a hard-to-maintain thing. Is there perhaps a coverage option to omit type hints related stuff?

JiriPavela (commented):
Good job, let's discuss the best course of action for some of the minor remarks and then merge it.

perun/check/integral_comparison.py
tfiedor: I think this might be automatic change by PyCharm. I see no reason for full path.

perun/check/linear_regression.py
tfiedor: Probaly IDE move.

perun/postprocess/normalizer/run.py
tfiedor: No, old change, when I just started with typing.

perun/postprocess/regression_analysis/data_provider.py
tfiedor: Same as above.

perun/utils/structs.py
tfiedor: Aggree, `object` is better.

pyproject.toml
tfiedor: I briefly googled, but found no good solution, so, we might add some remark, that this is not best solution. There is not much documentation, I think, regarding the typing and coverage. Naturally, we could write test, which would simply instantiate the class, but it is waste of time.

JiriPavela (approved):

PR #116, opened 2023-09-25
Remove the Codecov token from workflow.

JiriPavela:
Repository secrets are not working properly for GH Actions triggered by PRs from forked repositories. Although we might need to retry the Codecov upload from time to time due to an upload error, it is still the better solution than to hard-code the token or miss a coverage upload from new contributors.

Sources:
- https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954/15
- https://github.com/codecov/feedback/issues/126


tfiedor (approved):
Amazing!

PR #120, opened 2023-10-06
Move a helper function to prevent circular dependency

martymichal:
Since the `sanitize_filepart` function does not rely on anything in the `perun.profile` module, it can be safely moved out into a more neutral module.

I encountered the following traceback while playing around with Perun.

```
Traceback (most recent call last):                                                                                                                                                                                                                            
  File "/var/home/marty/.local/bin/perun", line 5, in <module>                                                                                                                                                                                                
    from perun.cli import launch_cli                                                                                                                                                                                                                          
  File "/var/home/marty/Documents/projects/perun/perun/cli.py", line 55, in <module>                                                                                                                                                                          
    import perun.fuzz.factory as fuzz                                                                                                                                                                                                                         
  File "/var/home/marty/Documents/projects/perun/perun/fuzz/factory.py", line 27, in <module>                                                                                                                                                                 
    import perun.fuzz.evaluate.by_perun as evaluate_workloads_by_perun                                                                                                                                                                                        
  File "/var/home/marty/Documents/projects/perun/perun/fuzz/evaluate/by_perun.py", line 8, in <module>                                                                                                                                                        
    import perun.check.factory as check                                                                                                                                                                                                                       
  File "/var/home/marty/Documents/projects/perun/perun/check/factory.py", line 15, in <module>                                                                                                                                                                
    from perun.profile.helpers import ProfileInfo                                                                                                                                                                                                             
  File "/var/home/marty/Documents/projects/perun/perun/profile/helpers.py", line 29, in <module>                                                                                                                                                              
    import perun.logic.store as store                                                                                                                                                                                                                         
  File "/var/home/marty/Documents/projects/perun/perun/logic/store.py", line 22, in <module>                                                                                                                                                                  
    from perun.profile.factory import Profile                                                                                                                                                                                                                 
  File "/var/home/marty/Documents/projects/perun/perun/profile/factory.py", line 25, in <module>                                                                                                                                                              
    from perun.check.general_detection import get_filtered_best_models_of                                                                                                                                                                                     
  File "/var/home/marty/Documents/projects/perun/perun/check/general_detection.py", line 24, in <module>                                                                                                                                                      
    import perun.check.fast_check as fast_check                                                                                                                                                                                                               
  File "/var/home/marty/Documents/projects/perun/perun/check/fast_check.py", line 13, in <module>                                                                                                                                                             
    import perun.logic.runner as runner                                                                                                                                                                                                                       
  File "/var/home/marty/Documents/projects/perun/perun/logic/runner.py", line 27, in <module>                                                                                                                                                                 
    import perun.collect.trace.optimizations.optimization                                                                                                                                                                                                     
  File "/var/home/marty/Documents/projects/perun/perun/collect/trace/optimizations/optimization.py", line 6, in <module>                                                                                                                                      
    from perun.profile.helpers import sanitize_filepart                                                                                                                                                                                                       
ImportError: cannot import name 'sanitize_filepart' from partially initialized module 'perun.profile.helpers' (most likely due to a circular import) (/var/home/marty/Documents/projects/perun/perun/profile/helpers.py)
```

### DIFF
--- a/perun/collect/trace/optimizations/optimization.py
+++ b/perun/collect/trace/optimizations/optimization.py
@@ -3,8 +3,7 @@
 
 import collections
 
-from perun.profile.helpers import sanitize_filepart
-from perun.utils.helpers import SuppressedExceptions
+from perun.utils.helpers import (SuppressedExceptions, sanitize_filepart)
 from perun.collect.trace.optimizations.structs import Optimizations, Pipeline, Parameters, \
     CallGraphTypes, ParametersManager, CGShapingMode
 import perun.collect.trace.optimizations.resources.manager as resources
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,4 +2,5 @@ ignore:
   - "perun/collect/trace"
   - "perun/thirdparty"
   - "tests/test_tracer"
+  - "tests/sources"
 
--- a/perun/logic/runner.py
+++ b/perun/logic/runner.py
@@ -469,7 +469,7 @@ def generate_jobs_on_current_working_dir(
     :param int number_of_jobs: number of jobs that will be run
     :return: status, generated profile, and associated job
     """
-    workload_generators_specs = workloads.load_generator_specifications()
+    workload_generators_specs: dict[str, GeneratorSpec] = workloads.load_generator_specifications()
 
     log.print_job_progress.current_job = 1
     collective_status = CollectStatus.OK
--- a/perun/logic/runner.py
+++ b/perun/logic/runner.py
@@ -6,7 +6,7 @@ import subprocess
 import signal
 import types
 
-from typing import Any, Iterable, Optional, TYPE_CHECKING, cast, Callable
+from typing import Any, Iterable, Optional, TYPE_CHECKING, cast, Callable, Type
 
 import distutils.util as dutils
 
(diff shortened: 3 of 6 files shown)


perun/collect/trace/optimizations/optimization.py
```diff
@@ -3,7 +3,7 @@
 
 import collections
 
-from perun.profile.helpers import sanitize_filepart
+from perun.utils.helpers import sanitize_filepart
```
JiriPavela: Seems like this import could be merged with the one below it, to keep the imports concise.

martymichal: Good catch!

JiriPavela (approved):
I fixed the new typing errors that emerged after a new mypy version has been released. I think we're ready to merge now.

martymichal:
I think the last two commits can be squashed.

JiriPavela:
Good idea, I think we could perhaps go even one step further and do squash&merge of the entire PR. What do you think @tfiedor?

tfiedor:
> Good idea, I think we could perhaps go even one step further and do squash&merge of the entire PR. What do you think @tfiedor?

I am not fan of squash, unless the commiter is dirty bastard (i.e., writing commit messages such as `asdf`, or `blabla`, or `MYNAME`). Which is not this case. 

EDIT: Also, you will appreciate more granular history once you bugfix some issue introduced in history, needing `git bisect` and finding precise chunk which caused error. So I would keep the history as it is, and rather focus on having green checkmarks for the PR.

martymichal:
I agree on not squashing the whole PR but the last two commits are related, so I don't see a problem with squashing those two (I regularly amend commits and then force push).

But every maintainer has a different taste, so you decide :).

tfiedor (approved):

PR #121, opened 2023-10-06
Remove demandimport module

martymichal:
The demandimport module relies on the imp module which has been deprecated in Python 3.4[0] and then removed in Python 3.12[1]. Remove it as it is no longer maintained (last commit in 2019), the performance benefit is marginal and optional imports can be managed in alternative ways (such as using try/expect or by utilising a build system).

[0] https://bugs.python.org/issue17177
[1] https://docs.python.org/dev/whatsnew/3.12.html#imp
[2] https://github.com/bwesterb/py-demandimport/commit/ca9c11635407b4b012c4b70f6fe03729b924eacd

### DIFF
--- a/perun/collect/trace/configuration.py
+++ b/perun/collect/trace/configuration.py
@@ -11,10 +11,7 @@ from perun.collect.trace.values import OutputHandling
 
 from perun.utils.exceptions import InvalidBinaryException
 from perun.utils import find_executable
-import perun.logic.temp as temp
-
-# Import on demand since eBPF support is optional
-import perun.collect.trace.ebpf.engine as bpf
+from perun.logic import temp
 
 
 class Configuration:
@@ -113,6 +110,8 @@ class Configuration:
         if self.engine == 'stap':
             self.engine = SystemTapEngine(self)
         else:
+            # Import on demand since eBPF support is optional
+            import perun.collect.trace.ebpf.engine as bpf
             self.engine = bpf.BpfEngine(self)
 
     def get_functions(self):
--- a/perun/collect/trace/configuration.py
+++ b/perun/collect/trace/configuration.py
@@ -3,7 +3,6 @@
 
 import os
 import time
-import demandimport
 
 from perun.collect.trace.collect_engine import CollectEngine
 from perun.collect.trace.systemtap.engine import SystemTapEngine
@@ -15,8 +14,7 @@ from perun.utils import find_executable
 import perun.logic.temp as temp
 
 # Import on demand since eBPF support is optional
-with demandimport.enabled():
-    import perun.collect.trace.ebpf.engine as bpf
+import perun.collect.trace.ebpf.engine as bpf
 
 
 class Configuration:
--- a/perun/fuzz/interpret.py
+++ b/perun/fuzz/interpret.py
@@ -1,12 +1,10 @@
 """ Module contains a set of functions for fuzzing results interpretation."""
 from __future__ import annotations
 
-import demandimport
 import os.path as path
 import difflib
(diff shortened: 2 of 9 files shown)


JiriPavela:
Thanks for the PR. We used `demandimport` not only for optional dependencies (for which try/except is sufficient), but also for lazy importing to improve Perun startup time. That being said, lazy imports were not solved systematically which explains the still too long startup time.

I was aware that demandimport is deprecated, however, we have not found a replacement for lazy imports yet. [PEP 690](https://peps.python.org/pep-0690/) seemed like it might solve our issue – until it was rejected. So for now, I think removing `demandimport` is the right move (to enable 3.12 support) and the next step should be finding a suitable alternative for lazy imports.

tfiedor:
> Thanks for the PR. We used `demandimport` not only for optional dependencies (for which try/except is sufficient), but also for lazy importing to improve Perun startup time. That being said, lazy imports were not solved systematically which explains the still too long startup time.
> 
> I was aware that demandimport is deprecated, however, we have not found a replacement for lazy imports yet. [PEP 690](https://peps.python.org/pep-0690/) seemed like it might solve our issue – until it was rejected. So for now, I think removing `demandimport` is the right move (to enable 3.12 support) and the next step should be finding a suitable alternative for lazy imports.

I believe, the main issue of the slow startup was `bokeh` and `bkcharts`, which we graciously removed, so we could look at the startup again.

martymichal:
It's better to have concrete values, so let's see.

Before this patch:

```
; time perun
real    0m3,811s
user    0m3,410s
sys     0m0,812s
```

[importtime-pre.txt](https://github.com/Perfexionists/perun/files/12922398/importtime-pre.txt)

After this patch:

```
; time perun
real    0m3,546s
user    0m3,301s
sys     0m0,804s
```

[importtime-post.txt](https://github.com/Perfexionists/perun/files/12922399/importtime-post.txt)

The star-up timing values need to be taken with a pinch of salt as they are dependent on so much but what they show is that removing `demandimport` does not cause a regression in the start-up time.

The `importtime` outputs are much more interesting, though. The biggest offenders seem to be:

- `angr` - ~1.3 second (`unicorn` takes up ~0.5 second of that mainly due to using `distutils` which pulls in `pkg_resources`).
- `matplotlib` - ~0.4 second
- `scipy` - ~0.4 second
- `pandas` - ~0.2 second
- `numpy` - ~0.1 second

Of these only `angr` seems to have some low-hanging fruit to boost import time.

JiriPavela:
Really interesting results, I believe that `demandimport` originally helped a bit with the startup time but it's quite possible that since it wasn't properly maintained since python 3.7, it might not be lazily importing in some scenarios anymore.

Let's get rid of `demandimport` for now and solve lazy imports later on (#127).

martymichal:
> I believe that `demandimport` originally helped a bit with the startup time but it's quite possible that since it wasn't properly maintained since python 3.7, it might not be lazily importing in some scenarios anymore.

I think it did that but the point of the lazy import is to do it uniformly. One stray import will regress the import time again. E.g., the only times this PR touches imports of `bokeh` is in `perun/utils/view_helpers.py` and `perun/view/scatter/factory.py` but it is also imported in `perun/utils/bokeh_helpers.py`. The same applies to `pandas` and possibly even the others.

tfiedor (approved):
The changes seems fine, I will look at the checks as well, before merging (and before Jirka adds his own review).

JiriPavela (approved):
I fixed the bcc import not being optional and the tests are passing now. I think we're ready for a merge.

PR #122, opened 2023-10-06
Update .gitignore

martymichal:
This is a combination of GitHub's .gitignore's for Python[0] and JetBrains[1] and a bit of Perun on top..

[0] https://github.com/github/gitignore/blob/main/Python.gitignore
[1] https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,183 @@
-*.pyc
-.idea/*
-.tox/*
+# Perun
+.perun/
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,240 @@
...
+#  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
+#  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
+#  and can be added to the global gitignore or merged into this file.  For a more nuclear
+#  option (not recommended) you can uncomment the following to ignore the entire idea folder.
+#.idea/
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# User-specific stuff
+.idea/**/workspace.xml
```
tfiedor: Please, lets merge this to `.idea/**` or `.idea/*`. I don't think there is valid reason for not ignoring anything related to IDEs.

tfiedor (changes requested):
I think the new `.gitignore` is good (I guess generated by some tool), but I would squash the idea directories (and maybe, if there are more similar cases that I overlooked, I would squash them as well; I'd rather keep the stuff short and concise).

.gitignore
JiriPavela: I agree. There is an explanation (in the long comment just above the shown snippet) of why the .idea ignores are scattered like that, but I think ignoring the entire .idea shouldn't really introduce any problem.

martymichal: Done.

tfiedor (approved):

JiriPavela (approved):

Issue #97, opened 2023-06-14
Repository maintenance


JiriPavela:
The Perun upstream repo needs a refresh in terms of its tooling, automation, CI/CD, etc.

Here's an list of tasks that should be completed (perhaps in this order to make things easier) as part of such refresh:

- [x] Remove the 'master' branch. (TF)
- [x] Setup GH Actions (auto tagging, documentation, changelog, upload to pip, ...). (TF?)
- [x] Move our project metadata and project installation from setup.py to pyproject.toml (#92 ). (JP)
- [x] Adopt the development task automation tool Tox (https://tox.wiki/en/latest/index.html, #95 ) . This should help us with the common issues of installing, testing and checking Perun (or commits) in local environment. (JP)
  - [x] Extend the Tox configuration with linting, doc generating and static type checking commands. (JP)
  - [x] Consider adopting a code formatter, e.g., [Black](https://github.com/psf/black) (pre-commit hook or code guidelines). (JP)
  - [x] Integrate Tox with GH actions (#95 ).
- [x] Add testing automation for newer Python versions. This should also make our future transitions to new Python versions a bit smoother (#94 ). (JP)
- [ ] Release Perun for Windows.
- [x] Make "Perferts" organization. (TF)
- [ ] Consider creating a Docker or Vagrant image to fully setup Perun environment, as we depend not only on Python dependencies, but some system packages or libraries as well.
- [x] Finish type hints. (TF)
- [ ] Reduce the number of dependencies. (JP, TF)
  - [x] Pruned obsolete or no longer needed dependencies.
  - [x] Analyze which dependencies are used very little and could be substituted with a local code.
- [ ] Create ReadTheDocs page for Perun.

PR #123, opened 2023-10-06
Format with black

martymichal:
Helps with #97.


JiriPavela:
Good job, please fix the typing errors first and then I will look into the individual files and see whether we need to custom-format (`# fmt: off`) some code constructs (typically collections of large number of constants, etc.).

JiriPavela (approved):
I polished the PR and I think we are ready for a merge. There might still be some code blocks that were perhaps formatted too aggressively, but we can solve those ad-hoc as we spot them.

Also, there are some Codacy issues that I believe were in the codebase before this PR, but they were likely triggered again by the formatting changes. I would leave fixing those issues for the #133.

docs/_static/templates/supported_module_names.py
```diff
@@ -1,12 +1,20 @@
...
+        )
     return {
-        'vcs': ['git'],
-        'collect': ['trace', 'memory', 'time'],
-        'postprocess': ['filter', 'normalizer', 'regression-analysis'],
-        'view': ['alloclist', 'bars', 'flamegraph', 'flow', 'heapmap', 'raw', 'scatter']
+        "vcs": ["git"],
+        "collect": ["trace", "memory", "time"],
+        "postprocess": ["filter", "normalizer", "regression-analysis"],
+        "view": [
```
tfiedor, line 11: I don't really like how black formatted this.

tfiedor (approved):

PR #124, opened 2023-10-06
Enable testing on Python 3.12

martymichal:
Python 3.12 was released a few days ago[0], so enabling testing on it is welcome.

[0] https://docs.python.org/dev/whatsnew/3.12.html

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
@@ -97,7 +97,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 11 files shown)


JiriPavela:
Ok, so although Perun codebase itself is now ready for 3.12, it seems like some of our dependencies not so much. The `capstone` python binding (used by `angr`) is still using distutils (https://github.com/capstone-engine/capstone/issues/2005). While we could make `angr` an on-demand import to bypass this import issue on startup, it is currently our only tool for extracting CG and CFG of profiled binaries. For this reason, I suggest we wait until all our important dependencies are ready for 3.12.

tfiedor:
> Ok, so although Perun codebase itself is now ready for 3.12, it seems like some of our dependencies not so much. The `capstone` python binding (used by `angr`) is still using distutils ([capstone-engine/capstone#2005](https://github.com/capstone-engine/capstone/issues/2005)). While we could make `angr` an on-demand import to bypass this import issue on startup, it is currently our only tool for extracting CG and CFG of profiled binaries. For this reason, I suggest we wait until all our important dependencies are ready for 3.12.

Indeed, I champion this suggestion. I don't think this is critical requirement (while it would be nice, we can wait few months, whether capstone developers will move forward with Python 3.12 support; if not we might fix it for them or circumvent the dependencies in some other way).

tfiedor (approved):

tfiedor:
I'm opening this issue and checking out if we can support py3.12 (it is really needed now).

tfiedor:
Good, I'm merging this, thank you Jirka and Ondra for making this work :-). We will need to apply Perun on Python 3.12. soon.

PR #134, opened 2023-10-27
Speed up the tests

tfiedor:
WIP


tfiedor:
I think, this is quite ready for review, it's harder and harder to find something to speed up. The complexity test could be definitely enhanced, but we could do it later. This now runs for about 2min for me, which is better (though I would be more happy with 1min). I am open for more suggestion how to speed this up. In the end, I made some speed-ups directly in Perun. Also I think, we could achieve some speed up, if we tested some functionality directly and not through whole integration tests executed by `invoke()` from click.

perun/cli.py
```diff
@@ -691,49 +701,58 @@ def collect(ctx: click.Context, **kwargs: Any) -> None:
...
               help='Strategy which determines how many mutations will be generated by certain'
               ' fuzzing rule in one iteration: unitary, proportional, probabilistic, mixed')
 @click.option('--regex-rules', '-r', nargs=1, required=False, multiple=True,
               callback=cli_helpers.single_yaml_param_callback, metavar='<file>',
-              help='Option for adding custom rules specified by regular expressions,'
+              help='Option for adding custom fuzzing rules specified by regular expressions,'
               ' written in YAML format file.')
 @click.option('--no-plotting', '-np', is_flag=True, required=False,
-              help='Avoiding sometimes lengthy plotting of graphs.')
+              help='Will not plot the interpretation of the ruzzing in form of graphs.')
```
JiriPavela: Seems like there is a typo in _fuzzing_?

perun/logic/config.py
```diff
@@ -122,6 +135,15 @@ def safe_get(self, key: str, default: Any) -> Any:
     def get(self, key: str) -> Any:
         """Returns the value of the key stored in the config.
 
+        :param str key: list of section separated by dots
+        :returns value: retrieved value of the key at config
+        :raises exceptions.MissingConfigSectionException: if the key is not present in the config
+        """
+        return self._get(key)
+
+    def _get(self, key):
```
JiriPavela: Looks like we are missing type hints here. They should perhaps be the same as in the above function `get`?

perun/logic/pcs.py
```diff
@@ -41,13 +41,14 @@ def get_path() -> str:
 
 
 @singleton
-def get_vcs_type() -> str:
-    """Returns the type of the wrapped version control system
+def get_vcs_type_and_url() -> tuple[str, str]:
+    """Returns the type and url of the wrapped version control system
 
     :return: type of the wrapped version control system
```
JiriPavela: Just a minor detail, the `:return:` string was not updated.

tests/test_cli.py
```diff
@@ -73,19 +73,18 @@ def test_cli(monkeypatch, pcs_full):
 
 def run_non_param_test(runner, test_params, expected_exit_code, expected_output):
     result = runner.invoke(cli.postprocessby, test_params)
+    print(test_params)
```
JiriPavela: Perhaps a helper `print()` that you forgot to delete?

tests/test_fuzzing.py
```diff
@@ -227,6 +251,8 @@ def test_fuzzing_hangs(pcs_full):
 
     hang_test = os.path.join(examples,  "hang-test", "hang")
 
+    # Fixme: This test is shaky, and should be implemented in different way; it can sometimes fail with error
+    # during the initial testing
```
JiriPavela, line 255: I think this should be extracted into a separate issue so that we know this needs our attention (tests that fail nondeterministically are really annoying).

JiriPavela (commented):
Really great job. I found some minor details, could you please address them? After that, I think we are ready for a merge.

JiriPavela (approved):

tfiedor:
I hath address thy issues. I'm merging.

PR #145, opened 2023-11-27
Port to Meson build system

### LINKED ISSUE #125: Adoption of a full-fledged build system
I spent some time thinking what is the best way to include https://gitlab.com/martymichal/sysrapl in Perun since it needs to build an eBPF module which is then utilized by a potential Python extension. 

Currently, Perun makes use of [setuptools](https://github.com/Perfexionists/perun/blob/devel/pyproject.toml#L3) that [has a way of including extensions](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html) but it is severely underpowered when compared to build systems with full support for the C/C++ ecosystem. I don't see a straightforward way of reproducing the [build steps](https://gitlab.com/martymichal/sysrapl/-/blob/main/src/meson.build) using setuptools without resorting to inherently fragile and unmaintainable spaghetti of Python code. Build systems cause a spaghetti of their own but they are meant for building projects unlike Python.

I spent some time searching and stumbled on a [large](https://github.com/python-cmaketools/python-cmaketools) [number](https://github.com/scikit-build/scikit-build) [of](https://github.com/mesonbuild/meson-python) [projects](https://github.com/scikit-build/scikit-build-core) that integrate to the "new build paradigm" of Python (as per [PEP 517](https://www.python.org/dev/peps/pep-0517/)) allowing to utilize build systems like CMake, Meson,.. while keeping the `pyproject.toml` file and only utilizing the build systems for more complex tasks (e.g., building C/C++ extensions). Some of these tools are still of pre-release quality (e.g., the `meson-python` project does not have support for optional dependencies -> `meson-python: error: Unsupported dynamic fields: "optional-dependencies"`) but others have seen more broad usage (e.g., https://github.com/scikit-build/scikit-build).

I still am unsure whether this is truly needed but my gut tells me that since perun touches low-level technologies like eBPF and wants to build tools utilizing it, having a full-fledged build system for the task seems like a good thing.

martymichal:
Powerful build system for building potential Python extensions and in
general any C/C++/<your-favourite-language> code.

Straightforward Python-like syntax, integration with editable
environments using automatic rebuilds, source/wheel distribution, build
introspection, reproducible builds and more!

The Makefile is still present as all interaction with Meson should be
done via pip/pypa-build and getting all the options correct is just too
much.

All references to setuptools are gone and instead there are new
dependencies on Meson, meson-python, ninja and patchelf. Meson being the
build system, meson-python the project linking the Python ecosystem with
Meson, ninja the build tool and patchelf a required dependency of
meson-python. All are easily accessible across platforms.

The best part of all this is that the only visible change is that you
need the new dependencies. The rest of the development process should be
the same as it was before the port.

Closes https://github.com/Perfexionists/perun/issues/125


meson.build
```diff
@@ -0,0 +1,30 @@
+project(
+    'perun',
+    'c',
+    version: '0.21.6',
```
JiriPavela: Note to @tfiedor: this might need updating based on merge order of this and #148 .

JiriPavela (approved):
Seems like everything works perfectly, thanks for solving this.

tfiedor (approved):
I approve this, looks good to me.

However, before I merge it:

1. I will first merge the refactorings,
2. I'm not sure if this will not break the release github action somehow, since master makefile was change quite a lot.
3. I need to clarify one things: in each meson.build one has to explicitly specify all the files? Is there some way of automating this? pyproject.toml/setup.py usually automatically collected all `.*py` files. This way, if I want to add new module to perun I will need to modify these `meson.build` files or create new ones. Is that right? This might be little unconvenient.

martymichal:
>  2. I'm not sure if this will not break the release github action
>     somehow, since master makefile was change quite a lot.

I can go through the workflows and double-check everything is alright.

>  3. I need to clarify one things: in each meson.build one has to
>     explicitly specify all the files? Is there some way of automating
>     this? pyproject.toml/setup.py usually automatically collected all
>     |.*py| files. This way, if I want to add new module to perun I will
>     need to modify these |meson.build| files or create new ones. Is that
>     right? This might be little unconvenient.

Explicit specification of files is tied to how Meson works. No room for 
unneeded files being processed during the build process (which sometimes 
happened if you had some files in the project not included in the 
.gitignore). Yes, it can be tedious to get started with a new module but 
before we merge this the templates could be updated to include the build 
files and update the existing ones.

Also, while an inconvenience, it's just a one of inconvenience per 
module and even large modules (like `tracer`) require e.g. 5 minutes to 
put together. Also, to make it a bit easier to include multiple files at 
once, one can also put together a little script that list all files in a 
directory and formats them in a way `meson.build` requires.

perun/collect/bounds/meson.build
```diff
@@ -0,0 +1,22 @@
...
+      perun_collect_bounds_files,
+      subdir: perun_collect_bounds_dir,
+  )
+
+  install_subdir(
+      'bin',
+      install_dir: py3.get_install_dir() / perun_collect_bounds_dir,
+      install_tag: 'python-runtime',
+  )
+  #endif
```
martymichal, line 22: While going through the code I stumbled on this one remnant. How should this be handled? At least on my machine the `bound` collector will not work due to the missing dependency. Because it is not bundled with Perun, it might be worthwhile to set the dependency as a required one but a better approach would probably be to bundle it or even better update the collector to support `libedit-3`.

Proper disabling of modules can be done but it will be much more involved than just not installing it.

martymichal:
>> 2. I'm not sure if this will not break the release github action somehow, since master makefile was change quite a lot.

> I can go through the workflows and double-check everything is alright.

I think all should be good. All Makefile rules exist, all do what they're supposed to, documentation is build in the same place and build artefacts are being built correctly in the right place.

I left one more comment about handling dependency on `libedit`. Otherwise this should be good to go.

meson.build
martymichal: Done.

perun/collect/bounds/meson.build
tfiedor: This is OK for now, though we should handle the bounds collector better, you are right.

tfiedor:
> > > 2. I'm not sure if this will not break the release github action somehow, since master makefile was change quite a lot.
> 
> > I can go through the workflows and double-check everything is alright.
> 
> I think all should be good. All Makefile rules exist, all do what they're supposed to, documentation is build in the same place and build artefacts are being built correctly in the right place.
> 
> I left one more comment about handling dependency on `libedit`. Otherwise this should be good to go.

I believe you it is working, however, I have quite experience with CI/CD and I know that even though stuff works locally or seems right, it tends to break. So I first want to merge the refactoring stuff, and then merge this and if something breaks, we will fix it immediately.

tfiedor:
> 2. I'm not sure if this will not break the release github action somehow, since master makefile was change quite a lot.
> I can go through the workflows and double-check everything is alright.
> 3. I need to clarify one things: in each meson.build one has to explicitly specify all the files? Is there some way of automating this? pyproject.toml/setup.py usually automatically collected all |.*py| files. This way, if I want to add new module to perun I will need to modify these |meson.build| files or create new ones. Is that right? This might be little unconvenient.
> Explicit specification of files is tied to how Meson works. No room for unneeded files being processed during the build process (which sometimes happened if you had some files in the project not included in the .gitignore). Yes, it can be tedious to get started with a new module but before we merge this the templates could be updated to include the build files and update the existing ones. Also, while an inconvenience, it's just a one of inconvenience per module and even large modules (like `tracer`) require e.g. 5 minutes to put together. Also, to make it a bit easier to include multiple files at once, one can also put together a little script that list all files in a directory and formats them in a way `meson.build` requires.

Can't you at least do some kind of wildcards? line `*.py`? That would help a lot, I think. Since creating packages is not so common and is usually done by me/Jirka or is supervised by us, hence we will make sure the `meson.build` is working, that would be OK with me. But adding each new module, or moving modules would be IMO quite tedious.

martymichal:
>Can't you at least do some kind of wildcards? line `*.py`? That would help a lot, I think. Since creating packages is not so common and is usually done by me/Jirka or is supervised by us, hence we will make sure the `meson.build` is working, that would be OK with me. But adding each new module, or moving modules would be IMO quite tedious.

The documentation of Meson talks about this:

- https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard
- https://mesonbuild.com/FAQ.html#but-i-really-want-to-use-wildcards

So, technically speaking the use of wildcards is possible but will cause different inconveniences and in general is discouraged.

tfiedor:
> Can't you at least do some kind of wildcards? line `*.py`? That would help a lot, I think. Since creating packages is not so common and is usually done by me/Jirka or is supervised by us, hence we will make sure the `meson.build` is working, that would be OK with me. But adding each new module, or moving modules would be IMO quite tedious.
> The documentation of Meson talks about this: - https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard - https://mesonbuild.com/FAQ.html#but-i-really-want-to-use-wildcards So, technically speaking the use of wildcards is possible but will cause different inconveniences and in general is discouraged.

@JiriPavela what is your take on this? I mean I get the reasons, and while I would rather like autodiscovery of modules/sources, I can live with explicit writing.

tfiedor:
We discussed this with Jirka during a call, and we aggreed we will move forward with this change. Well done, @HarryMichal there might be some stuttering in future, but it will definitely help us in different aspects (integrating new profilers) and we believe we will need this in near future. I will not make new release anytime soon though (after several more refactoring) so we will see, if this will break the release action that deploys perun to pypi (but I'm actually fairly certain it will work finely).

PR #147, opened 2023-12-18
Fix issue with updated termcolor

tfiedor:
This PR: 
  * bumps termcolor version;
  * adds better typing for colours (used by termcolor).

This basically fixes an issue with failing mypy.


JiriPavela (approved):
Good, let's merge it.

PR #149, opened 2024-01-02
Winter refactoring: Refactor formatting strings.

tfiedor:
This pull request is part of winter refactoring.

1. It transforms most of `.format()` to `f""` formatting string notation;
2. It removes most of the `print` statements in favour of `log.info`;
3. It polishes some minor things.

### DIFF
--- a/perun/logic/config.py
+++ b/perun/logic/config.py
@@ -173,7 +173,6 @@ def write_config_to(path: str, config_data: dict[str, Any]) -> None:
     :param str path: path where the config will be stored to
     :param dict config_data: dictionary with contents of the configuration
     """
-    perun_log.msg_to_stdout("Writing config '{}' at {}".format(config_data, path), 2)
     with open(path, "w") as yaml_file:
         YAML().dump(config_data, yaml_file)
 
@@ -189,7 +188,7 @@ def read_config_from(path: str) -> dict[str, Any]:
         return streams.safely_load_yaml_from_file(path)
     except scanner.ScannerError as scanner_error:
         perun_log.error(
-            "corrupted configuration file '{}': {}\n".format(path, str(scanner_error))
+            f"corrupted configuration file '{path}': {scanner_error}\n"
             + "\nPerhaps you did not escape strings with special characters in quotes?"
         )
         return {}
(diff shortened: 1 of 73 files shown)


perun/logic/config.py
```diff
@@ -173,7 +173,6 @@ def write_config_to(path: str, config_data: dict[str, Any]) -> None:
     :param str path: path where the config will be stored to
     :param dict config_data: dictionary with contents of the configuration
     """
-    perun_log.msg_to_stdout("Writing config '{}' at {}".format(config_data, path), 2)
```
JiriPavela, line 176: There seems to be no replacement for this message. Is this intended or just a mistake?

perun/logic/index.py
```diff
@@ -448,11 +445,8 @@ def write_entry_to_index(index_file: str, file_entry: BasicIndexEntry) -> None:
                     looked_up_entry.path == file_entry.path
                     and looked_up_entry.time == file_entry.time
                 ):
-                    perun_log.msg_to_stdout(
-                        "{0.path} ({0.time}) already registered in {1}".format(
-                            file_entry, index_file
-                        ),
-                        0,
+                    perun_log.warn(
+                        "{file_entry.path} ({file_entry.time}) already registered in {index_file}",
```
JiriPavela: Seems like this should be a f-string.

perun/view/raw/run.py
```diff
@@ -26,22 +26,19 @@ def show(profile: Profile, **kwargs: Any) -> None:
...
-        print(log.in_color("collector:", RAW_KEY_COLOUR))
+        log.info(log.in_color("collector:", RAW_KEY_COLOUR))
         collector_info = profile["collector_info"]
         for collector_item in ["name", "params"]:
             if collector_item in collector_info.keys():
-                print(
+                log.info(
                     int(raw_indent) * 1 * " "
                     + "- {}: {}".format(
                         log.in_color(collector_item, RAW_KEY_COLOUR),
```
JiriPavela, line 44: Seems like one `.format` string got through here.

JiriPavela (approved):
Good job, the codebase seems to be a bit more readable (and should be faster a bit as well) now when f-strings are used instead of `format`. Please look into the minor remarks I added in the review before merging.

perun/logic/config.py
tfiedor: Yes, it was intended.

perun/view/raw/run.py
tfiedor: Intended.

PR #151, opened 2024-01-10
Winter refactoring: refactor imports

tfiedor:
This PR:

1. Reduces some imports;
2. Unifies some imports;
3. Moves some stuff around;
4. Fixes some grammatical errors/issues.


perun/check/exclusive_time_outliers.py
```diff
@@ -65,20 +65,19 @@
...
 
-from typing import Optional, Iterable, Any
-from difflib import get_close_matches
-
-import pandas as pd
+import difflib
 import numpy as np
-from scipy import stats
+import pandas as pd
+import scipy.stats as stats
```
JiriPavela: I'm fairly certain that the change of `scipy.stats` import will now trigger a pylint refactor suggestion [R0402](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/consider-using-from-import.html). How should we handle this specific suggestion going forward? Should we add it to the pylint list of ignored warnings / suggestions?

perun/cli.py
```diff
@@ -36,42 +36,34 @@
 Graphical User Interface is currently in development and hopefully will extend
 the flexibility of Perun's usage.
 
-.. _Click: http://click.pocoo.org/5/
+.. _Click: https://click.pocoo.org/5/
```
JiriPavela: The link seems to be dead. Do we need it here? If so, I propose a change to https://click.palletsprojects.com/en/latest/ (it should always point to the most recent development click version).

perun/cli_groups/utils_cli.py
```diff
@@ -77,7 +74,7 @@ def create(template_type: str, **kwargs: Any) -> None:
     Unless ``--no-edit`` is set, after the successful creation of the files, an external editor,
     which is specified by :ckey:`general.editor` configuration key.
 
-    .. _jinja: http://jinja2.pocoo.org/
+    .. _jinja: https://jinja2.pocoo.org/
```
JiriPavela: Same as above. Current link seems to be https://jinja.palletsprojects.com/en/latest/ .

perun/collect/bounds/run.py
```diff
@@ -6,19 +6,19 @@
...
+import click
 import os
 import shutil
 import time as systime
 
 from subprocess import SubprocessError
 from typing import Any
 
-import click
-
```
JiriPavela: Is there a reason for moving the import around? This ordering will violate PEP8 import guidelines (https://peps.python.org/pep-0008/#imports) and will trigger pylint [C0411](https://pylint.pycqa.org/en/v2.17.7/user_guide/messages/convention/wrong-import-order.html).

perun/collect/complexity/run.py
```diff
@@ -7,22 +7,19 @@
 from __future__ import annotations
 
 import dataclasses
+import click
```
JiriPavela, line 10: The same issue as above.

perun/collect/trace/systemtap/engine.py
```diff
@@ -27,11 +27,11 @@
 
 import perun.utils as utils
 import perun.utils.metrics as metrics
-from perun.utils.helpers import SuppressedExceptions
 from perun.logic.locks import LockType, ResourceLock, get_active_locks_for
 from perun.utils.exceptions import (
     SystemTapStartupException,
     SystemTapScriptCompilationException,
+    SuppressedExceptions,
```
JiriPavela, line 34: Maybe an idea for a separate issue: we should be careful with using `SuppressedExceptions` in loops and hot paths because a new context manager is created each time the code is executed. This can lead to a serious slowdown compared to the built-in try/except.

perun/logic/config.py
```diff
@@ -12,20 +12,17 @@
 import dataclasses
 import os
 import re
+import ruamel.yaml.comments as comments
 import sys
 
 from ruamel.yaml import YAML, scanner
-import ruamel.yaml.comments as comments
```
JiriPavela: Why move this import? Similar issue as with click.

perun/logic/config_templates.py
```diff
@@ -54,13 +54,12 @@
...
 from __future__ import annotations
 
+import jinja2
 import os
 import subprocess
 
 from typing import Iterable, Any
 
-import jinja2
-
```
JiriPavela: Same as above.

JiriPavela (approved):
Thanks for the PR. My only issue with this is that the ordering scheme of imports (by type of import + alphabetically if I understand it correctly) violates PEP8 and will cause numerous pylint refactor/convention warnings. We should probably discuss a consistent importing scheme for this codebase.

perun/collect/trace/systemtap/engine.py
tfiedor: Can you create an issue?

tfiedor:
I have reordered all of the imports to three groups: (1) standard, (2) thirdparty, (3) perun. And I have fixed the broken links. I will merge this soon.

PR #155, opened 2024-01-28
Refactor selection mechanism of Perun

tfiedor:
This PR extracts the functionality of selecting distinct versions or profiles.

In particular:
  1. It provides abstract class for handling the selection of version/profiles.
  2. It provides basic implementation in form of WholeRepositorySelection
  3. Add selection factory to pcs module.


tests/test_select.py
```diff
@@ -0,0 +1,59 @@
...
+import git
+import pytest
+
+# Third-Party Imports
+
+# Perun Imports
+from perun import vcs
+from perun.profile import helpers as profile_helpers
+from perun.select.abstract_base_selection import AbstractBaseSelection
+from perun.select.whole_repository_selection import WholeRepositorySelection
```
JiriPavela, line 13: Could we perhaps import both in `select.__init__.py` so that we can import directly from `perun.select`?

JiriPavela (approved):
Good, I like this solution, much easier to extend in the (near) future.

tests/test_select.py
tfiedor: Will be resolved in next PR.

PR #158, opened 2024-02-01
Rework the job arguments

tfiedor:
This PR:

1. Removes `args` from existence, and merge their functionality into `cmd`

### DIFF
--- a/tests/test_fuzzing.py
+++ b/tests/test_fuzzing.py
@@ -335,7 +335,6 @@ def test_fuzzing_incorrect(pcs_with_root):
     result = runner.invoke(
         cli.fuzz_cmd,
         [
-            "--args", "-al",
             "--output-dir", ".",
             "--input-sample", ".",
         ],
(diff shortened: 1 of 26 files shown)


tests/test_fuzzing.py
```diff
@@ -348,7 +347,6 @@ def test_fuzzing_incorrect(pcs_with_root):
         cli.fuzz_cmd,
         [
             "--cmd", "ls",
-            "--args", "-al",
             "--output-dir", ".",
         ],
     )  # fmt: skip
```
JiriPavela, line 352: Shouldn't the `--args` value be appended to the `--cmd`, or doesn't it matter for those tests?

JiriPavela (approved):
Ok, let's hope we won't later discover we actually need the `--args`.

tfiedor:
> Ok, let's hope we won't later discover we actually need the `--args`.

No we will not, and `cmd` is fine substitution. I think in future we might rather go for multiple (named) job matrixes instead (I wanted just some quick refactoring, so I omited the refactoring of job matrix).

tests/test_fuzzing.py
tfiedor: Good point, but it should not matter.

PR #177, opened 2024-02-16
Add option for user to initialize perun and git if Perun is run from outside of perun scope

### LINKED ISSUE #171: Rework the handling of NotInPerunRepository exception
Currently, when `NotInPerunRepository` exception is fired, the execution stops and user has to manually set up the Perun.

The functionality should be changed into following:
  1. When `.perun` does not exist and perun is run, then warning is issued to user and user is prompted whether user wants to initialize perun repository. If not confirmed the execution fails, otherwise, the following can happen:
     1. `.git` exists in the directory, then Perun is initialized with the git
     2. `.git` does not exist in the directory, then Perun is initilized with lightweight custom kernelvcs/osvcs, which is only based on OS/environment version.

tfiedor:
This PR partly reworks the way NotPerunRepositoryException is handled for CLI commands.

If the exception is raised, then the user is prompted whether he/she/they wants to initialize `.perun` together with `.git`. If there is no `.git` in the current working directory, an empty root commit is created (using `git commit --allow-empty -m "root"`). This is temporary solution, which will be replaced by lightweight VCS, that will be developed in near future.

Also this PR rework or fixes:

1. Fixes nondeterministic bug in tests (due to setting no paging)
2. Extracts confirmations from user to isolate function
3. Adds support for suppresing user prompts by `--say-yes` option.
4. Fixes minor issue in Logger.

### DIFF
--- a/perun/cli.py
+++ b/perun/cli.py
@@ -170,7 +170,7 @@ def cli(
     # through --no-pager set by default to False you enable the paging
     global DEV_MODE
     DEV_MODE = dev_mode
-    perun_log.ALWAYS_CONFIRM = say_yes
+    common_kit.ALWAYS_CONFIRM = say_yes
     perun_log.SUPPRESS_PAGING = no_pager
     perun_log.COLOR_OUTPUT = not no_color
 
--- a/perun/cli.py
+++ b/perun/cli.py
@@ -52,7 +52,7 @@ import click
 from perun.cli_groups import check_cli, config_cli, run_cli, utils_cli
 from perun.logic import commands, pcs, config as perun_config
 from perun.utils import exceptions, log as perun_log
-from perun.utils.common import cli_kit
+from perun.utils.common import cli_kit, common_kit
 from perun.utils.external import commands as external_commands
 from perun.collect.trace.optimizations.structs import (
     Pipeline,
@@ -65,7 +65,6 @@ from perun.utils.exceptions import (
     UnsupportedModuleException,
     NotPerunRepositoryException,
     IncorrectProfileFormatException,
-    EntryNotFoundException,
     MissingConfigSectionException,
     ExternalEditorErrorException,
 )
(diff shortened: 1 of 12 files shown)


perun/cli.py
```diff
@@ -105,6 +104,9 @@
     is_flag=True,
     help="Disables the colored output.",
 )
+@click.option(
+    "--say-yes", "-y", default=False, is_flag=True, help="Says yes to every confirmation prompt"
+)
```
JiriPavela, line 109: Just a small detail, there seems to be a missing dot at the end (all the other help strings have it).

perun/logic/commands.py
```diff
@@ -166,6 +168,42 @@ def init_perun_at(
...
+                    "Creating empty commit",
+                    status=perun_log.cmd_style('git commit --allow-empty -m "root"'),
+                )
+                try:
+                    external_commands.run_safely_external_command(
+                        'git commit --allow-empty -m "root"'
+                    )
+                    perun_log.minor_success("Creating empty commit")
+                except subprocess.CalledProcessError:
+                    perun_log.minor_fail("Creating empty commit")
```
JiriPavela, line 201: I know you always wanted to isolate all subprocess-related stuff to just one module. Should we perhaps import the currently used subprocess exceptions/errors in such module (I am not sure which module it is right now) and then internally import it from there?

JiriPavela (approved):

perun/logic/commands.py
tfiedor: Hm, good point, we could create an issue about this. I think my point was mainly (1) codacy is really annoying about this, (2) having as few `import subprocess` forces other users to use our safe or well-established functions in `perun.utils.external`. So far most of the usages left should be importing `CalledProcessorError`, which is okayish I guess, but it might be better to import them either to `perun.utils.external` (`__init__.py`) or to `perun.utils.exceptions`.

perun/cli.py
tfiedor: Good catch, however, I will not fix it, as I have a plan for rewritting/reworking the help messages (and maybe CLI commands as well). I want to unify it mostly. Btw. I have already asked AI to rewrite some of the messages, however, I was not really happy with it, so I will rewrite it myself manually, probably during friday/next weekend.

PR #180, opened 2024-02-19
Stop using distutils

martymichal:
Helps with support for Python 3.12.


martymichal:
Only now noticed this is already covered by https://github.com/Perfexionists/perun/pull/124/commits/97e65f6a188a12fffe42b9849d6223e57920ac7b. Closing.

PR #184, opened 2024-03-15
Add differential views of pairs of profiles.

tfiedor:
This PR adds new mode of perun, that visualizes the differences between pairs of profiles.

In particular it adds three new views:

1. Tabular view (`table`)
2. HTML Report with DataTables.js
3. Flamegraph Report.


perun/view_diff/table/__init__.py
JiriPavela, line 1: Should we perhaps add some docstring?

JiriPavela (approved):

PR #189, opened 2024-05-06
Add incremental sankey diff view

tfiedor:
This PR:

1. Fixes minor issues,
2. Makes flamegraph more versatile,
3. Makes diff flamegraph to support other resource types,
4. Improve diff report (add filtering per clusters or per type, filtering of values),
5. Add two versions of sankey graphs,
6. Fix minor issues in traces.


perun/profile/convert.py
```diff
@@ -157,13 +161,17 @@ def to_uid(record: dict[str, Any] | str) -> str:
         return to_string_line(record)
 
 
-def to_string_line(frame: dict[str, Any]) -> str:
+def to_string_line(frame: dict[str, Any] | str) -> str:
     """Create string representing call stack's frame
 
+    TODO: remake
```
JiriPavela: Is the TODO still relevant or did this change resolve the TODO?

perun/utils/common/common_kit.py
```diff
@@ -11,6 +14,7 @@
 import signal
 
 # Third-Party Imports
+import array
```
JiriPavela: I believe `array` is part of the Python standard library? https://docs.python.org/3/library/array.html

perun/utils/common/common_kit.py
```diff
@@ -437,4 +451,97 @@ def get_module(module_name: str) -> types.ModuleType:
...
+    :return: object
+    """
+    try:
+        yield disposable
+    except Exception:
+        # Re-raise the encountered exception
+        raise
+    finally:
+        del disposable
+        gc.collect()
```
JiriPavela, line 512: Neat, just curious, isn't there a heavy performance penalty for calling GC? Maybe the docstring should mention the potential overhead so that others don't overuse it.

perun/utils/common/common_kit.py
```diff
@@ -437,4 +451,97 @@ def get_module(module_name: str) -> types.ModuleType:
...
+
+
+def binary_search(
+    values: list[Any], value: Any, low: int, high: int, key: Callable[[Any], Any] = lambda x: x
+) -> int:
+    """Classical binary search algorithm
```
JiriPavela, line 518: Could maybe https://docs.python.org/3/library/bisect.html be used here? I'm not sure whether the interface is compatible with what you need.

perun/utils/common/common_kit.py
```diff
@@ -437,4 +451,97 @@ def get_module(module_name: str) -> types.ModuleType:
...
+        elif value > values[mid]:
+            low = mid + 1
+        else:
+            high = mid - 1
+    return low
+
+
+def add_to_sorted(
+    values: list[Any], value: Any, key: Callable[[Any], Any] = lambda x: x, max_pick: int = 0
+) -> None:
```
JiriPavela, line 543: Same, the bisect module should be able to do this as well?

perun/view_diff/sankey/run.py
```diff
@@ -0,0 +1,591 @@
...
+
+"""
+from __future__ import annotations
+
+# Standard Imports
+from typing import Any, Literal
+
+# Third-Party Imports
+from collections import defaultdict
+from dataclasses import dataclass
```
JiriPavela: Aren't these two packages part of the standard library?

perun/view_diff/sankey_incremental/run.py
```diff
@@ -0,0 +1,674 @@
...
+
+# Standard Imports
+from operator import itemgetter
+from typing import Any, Literal, Type, Callable
+from collections import defaultdict
+from dataclasses import dataclass
+import sys
+
+# Third-Party Imports
+import array
```
JiriPavela: Again, `array` should be part of the standard imports group I believe?

perun/view_diff/sankey_incremental/run.py
```diff
@@ -0,0 +1,674 @@
...
+    Increase: str = "rgba(255, 0, 0, 0.7)"
+    Decrease: str = "rgba(0, 255, 0, 0.7)"
+    Equal: str = "rgba(0, 0, 255, 0.7)"
+    DarkTarget: str = "rgba(255, 201, 74, 1)"
+    DarkBaseline: str = "rgba(49, 48, 77, 1)"
+    DarkIncrease: str = "#ea5545"
+    DarkDecrease: str = "#87bc45"
+    DarkEqual: str = "#27aeef"
+    Highlight: str = "rgba(0, 0, 0, 0.7)"
+    NoHighlight: str = "rgba(0, 0, 0, 0.2)"
```
JiriPavela: Doesn't the `sankey/run.py` define a ColorPalette class as well? It seems at least some colors are the same? Maybe consider merging the classes.

JiriPavela (approved):

perun/profile/convert.py
tfiedor: I guess it is resolved, but could be better.

perun/utils/common/common_kit.py
tfiedor: Good Idea to extend it with docstring. Just little anecdote: I worked with some huge json files that I was converting to pandas.DataFrame and then concatenating into single one; the explicit call to gc actually made it feasiable, as it simply refused to collect the garbage in between the loading (it was in for loop). So manual GC can definitely help, but needs some tweaks.

perun/utils/common/common_kit.py
tfiedor: This is neat, I actually didn't assume there would be such algorithm in standard library, and it should be usable here :)

perun/utils/common/common_kit.py
tfiedor: I don't think I can fully replace this function, since I want to keep only `max_pick` elements in the list, however, it could be extracted to higher function. Though I will replace the binary search + insertion with insort (unless it breaks something)

perun/view_diff/sankey_incremental/run.py
tfiedor: Good point, I actually want to extract all colours to isolate package, but for now, I extracted this to WebColorPalette class in perun/utils/structs.py

PR #202, opened 2024-06-17
TypeScript Web module

TomValent:
Can you finish the PR instead of me please? Tests and check throw some errors, but some are not from this module.


JiriPavela:
Thanks for opening the PR. @tfiedor and I will have a look at it and polish it before merging it into the upstream.

PR #213, opened 2024-06-23
Stand alone delta debugging command

MozartPanda:
This pull request includes the implementation of delta debugging for a stand alone cli command.


tfiedor:
Thanks for the PR, we are currently quite swamped with Jirka (we are leading some workshop tomorrow), hopefully we will look at it at the end of the week.

perun/cli.py
```diff
@@ -1249,6 +1250,43 @@ def fuzz_cmd(cmd: str, **kwargs: Any) -> None:
...
+    ),
+)
+@click.option(
+    "--output-dir",
+    "-o",
+    nargs=1,
+    required=True,
+    type=click.Path(exists=True, writable=True),
+    metavar="<path>",
+    help="The path to the directory where generated outputs will be stored.",
```
JiriPavela: Since all of these click options have `required=True`, I would consider turning the options into positional `click.argument`s instead. I noticed some of our click commands have options with `required=True` as well, but these will likely be refactored out in the future.

perun/cli.py
```diff
@@ -1249,6 +1250,43 @@ def fuzz_cmd(cmd: str, **kwargs: Any) -> None:
...
+    "--timeout",
+    "-t",
+    nargs=1,
+    required=False,
+    default=1.0,
+    type=click.FloatRange(0.001, None, False),
+    metavar="<float>",
+    help="Time limit for delta debugging (in seconds). Default value is 1s.",
+)
+def deltadebugging(cmd: str, **kwargs: Any):
```
JiriPavela: There is a missing return type hint, which should be `None` I believe.

perun/deltadebugging/factory.py
```diff
@@ -0,0 +1,77 @@
...
+    from perun.utils.structs import Executable
+
+def run_delta_debugging_for_command(
+    executable: Executable,
+    input_sample: list[str],
+    **kwargs: Any,
+) -> None:
+
+    timeout: float = kwargs.get("timeout", 1.0)
+    output_dir: str = os.path.abspath(kwargs["output_dir"])
```
JiriPavela: We are planning to replace all (or most) occurrences of `os.path` with `pathlib` equivalent constructions (https://docs.python.org/3/library/pathlib.html) in the future. Consider switching to the `pathlib` in the delta debugging package.

perun/deltadebugging/factory.py
```diff
@@ -0,0 +1,77 @@
...
+        if not program_fails:
+            if n == len(inp):
+                break
+            n = min(n * 2, len(inp))
+
+    log.minor_info('shortest failing input = ' + inp)
+    return create_debugging_file(output_dir, input_sample, inp)
+
+
+def readInput(input_file):
```
JiriPavela: This function is missing type annotations.

perun/deltadebugging/factory.py
```diff
@@ -0,0 +1,77 @@
...
+    if os.path.isfile(input_file):
+        with open(input_file, 'r') as file:
+            input_value = file.read()
+    else:
+        input_value = input_file
+
+    return input_value
+
+def create_debugging_file(output_dir, file_name, input_data):
```
JiriPavela: Here is another function without type annotations.

perun/deltadebugging/factory.py
```diff
@@ -0,0 +1,77 @@
...
+            complement = (inp[:int(start)] + inp[int(start + subset_length):])
+            try:
+                full_cmd = f"{executable} {complement}"
+                external_commands.run_safely_external_command(full_cmd, True, True, timeout)
+
+            except subprocess.TimeoutExpired:
+                inp = complement
+                n = max(n - 1, 2)
+                program_fails = True
+                print(complement)
```
JiriPavela: Ideally, try to use the appropriate method (warn, error, etc.) from the `utils.log` module instead of a plain print. It adds an annotation to the output ("Error: ..." or "Warning: ...") and helps the user with an estimation of the message severity.

perun/deltadebugging/factory.py
```diff
@@ -0,0 +1,77 @@
...
+
+def create_debugging_file(output_dir, file_name, input_data):
+    dir_name = 'delta_debugging'
+    full_dir_path = os.path.join(output_dir, dir_name)
+
+    file_dir = os.path.dirname(file_name)
+    if file_dir:
+        full_dir_path = os.path.join(full_dir_path, file_dir)
+
+    os.makedirs(full_dir_path, exist_ok=True)
```
JiriPavela: You can use the `touch_dir` function in `utils.common.common_kit` that is a wrapper over `os.makedirs`.

JiriPavela (commented):
Overall, the code looks good. Please look into the feedback I've provided and try to resolve it.

When writing the type annotations, you can check whether the annotations are correct by running `make check`, which does type checking using mypy.

Also, consider adding some tests for your new package (see the `tests` folder in `perun`) so that we know when something breaks in the future.

JiriPavela (approved):
Great, I think we're ready for a merge. Good job.

PR #216, opened 2024-07-02
Add SVS version control system

tfiedor:
This PR adds SVS version control, that is made as a default:

1. SVS has single branch and single version, that tracks nothing.
2. SVS is made default and is intented for gitless usage and version-less usage of perun.

### DIFF
--- a/perun/vcs/svs_repository.py
+++ b/perun/vcs/svs_repository.py
@@ -1,0 +1,103 @@
+"""SVS is a dummy Single Version System, which tracks the results as a single version, nothing more
+
+Contains concrete implementation of the functions needed by Perun to work without any version control system,
+like GIT, SVN, or other.
+"""
+
+from __future__ import annotations
+
+# Standard Imports
+from typing import Any, Iterator
+import getpass
+import os
+import socket
+import sys
+import time
+
+
+# Third-Party Imports
+
+# Perun Imports
+from perun.utils.structs import MajorVersion, MinorVersion
+from perun.vcs.abstract_repository import AbstractRepository
+
+SINGLE_VERSION_BRANCH: str = "master"
+SINGLE_VERSION_TAG: str = "HEAD"
+
+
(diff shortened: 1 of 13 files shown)


perun/vcs/svs_repository.py
```diff
@@ -0,0 +1,169 @@
...
+
+def get_time_of_creation(path: str) -> str:
+    """Returns the time of the creation of the path
+
+    :param path: path we are analysing the time of creation
+    :return: time of creation of the path
+    """
+    return time.ctime(
+        os.path.getctime(path) if sys.platform.startswith("win") else os.stat(path).st_ctime
+    )
```
JiriPavela, line 36: Looking at the library functions, it seems this function returns different information for Linux and Windows machines. For Windows, the `getctime` function will return the creation time of the path, while the `st_ctime` function will return the most recent metadata change of the path. Seems like the `st_birthtime` might be the correct substitution for at least some Linux machines, provided it works?

Also, I'd consider using `pathlib` instead of the `os.path` module. Obtaining the creation time of a path for different platforms could be done using `Path.stat()` and simply accessing different `stat_result` attributes according to the platform.

Finally, should we maybe move platform-dependent utilities into a separate module, similarly to how we bundle all `subprocess` methods in one place? The os-based library methods and attributes change quite a lot so having a single module or package should make it easier to maintain.

JiriPavela (approved):

perun/vcs/svs_repository.py
tfiedor: 1. Good point, I think the datetime is quite "useless" there, I just didn't want to output nothing when `perun status` is run. I initially had the birthtime, but typing was screaming, so I refactored it despite the "incorrectness".
2. I didn't use the pathlib, as I you wanted to refactor the paths anyways.
3. Good point, make a note somewhere. I think we should change it in separate PR.

Issue #203, opened 2024-06-19
Numpy 2.0 - breaking changes


PeterMocary:
With the release of NumPy 2.0 (16.6.2024), there are several breaking changes that need to be addressed. Since the migration might require some time, we should limit the version (``numpy>=1.24, <2.0``) for the time being.

Useful links:
- [NumPy 2.0 release notes](https://numpy.org/doc/stable/release/2.0.0-notes.html)
- [Numpy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html)

Triggered for example here:
```
Traceback (most recent call last):
[...]
File "[...]/perun/postprocess/regressogram/methods.py", line 12, in <module>
    import numpy.lib.histograms as numpy_bucket_selectors
ModuleNotFoundError: No module named 'numpy.lib.histograms'

---- Interactive python interpreter ----
Python 3.11.9 (main, Apr 17 2024, 00:00:00) [GCC 13.2.1 20240316 (Red Hat 13.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy.lib.histograms
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'numpy.lib.histograms'
>>> import numpy
>>> numpy.lib.histograms
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "[...]/lib64/python3.11/site-packages/numpy/lib/__init__.py", line 77, in __getattr__
    raise AttributeError(
AttributeError: numpy.lib.histograms is now private. If you are using a public function, it should be available 
in the main numpy namespace, otherwise check the NumPy 2.0 migration guide.
```

PeterMocary:
Temporary fix merged in PR #205.

Issue #106, opened 2023-08-28
Improve SciPy's integration precision (Python 3.9+)


JiriPavela:
Starting from SciPy 1.11.0, the [simpson](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.simpson.html#scipy.integrate.simpson) sampled integration implements a [more precise technique for odd number of intervals (even number of points)](https://docs.scipy.org/doc/scipy/release/1.11.0-notes.html#scipy-integrate-improvements). However, SciPy 1.11.0 dropped support for Python 3.8 and the change is not backported to 1.10.0.

Consider switching to this improved approximation once we drop 3.8 support as well (or add a case-split based on the interpreter version within the code). The integration is used in these two following modules:

https://github.com/Perfexionists/perun/blob/51593d9c2fd37b70ae8ba959e880d4a0db804204/perun/check/integral_comparison.py#L54

https://github.com/Perfexionists/perun/blob/51593d9c2fd37b70ae8ba959e880d4a0db804204/perun/check/local_statistics.py#L92

and should affect only this particular test (meaning some reference results will have to be updated):

https://github.com/Perfexionists/perun/blob/51593d9c2fd37b70ae8ba959e880d4a0db804204/tests/test_detection_methods.py#L93-L99

PR #221, opened 2024-07-04
Bump dependencies for Python 3.9 and fix related dependency issues

JiriPavela:
This PR bumps our dependencies to their latest versions that still support 3.9, while resolving some of our existing dependency issues, namely #106 and #203.


tfiedor (approved):
Seems fine to me.

Please, Jirka, write to Jirka H. and test if this breaks anything in RHEL10 or RHEL09.

Feel free to merge it.

PR #235, opened 2024-07-17
Fix jinja2 issue with editable build

### LINKED ISSUE #204: Jinja2 - PackageLoader unable to find the installation of perun
When loading a Jinja2 template the``PackageLoader`` is used to locate the templates within the package. When is the perun package installed using ``make dev`` the ``PackageLoader`` fails and resutlts in a crash. 

The ``PackageLoader`` is used in [view_diff](https://github.com/Perfexionists/perun/blame/eba7943c61c41dc3d7074dba6531bd35b8cfda4b/perun/view_diff/flamegraph/run.py#L196), [ktrace](https://github.com/Perfexionists/perun/blob/4481fba794f6e188ee113c072b657c783df30603/perun/collect/ktrace/bpfgen.py#L24), and in the [PIN engine](https://github.com/PeterMocary/perun/blob/067e3e5db9d665e6d29d5f1be6a28e6831a1509d/perun/collect/trace/pin/engine.py#L241), therefore this should be investigated further to ensure stable development enviroment.

Triggered on PIN engine when generating a pintool from templates:
```
[ERROR] while collecting by trace: error while collect phase: The 'perun' package was not installed in a way that PackageLoader understands.
  File "[...]/perun/perun/logic/runner.py", line 217, in run_phase_function
    phase_result = phase_function(**report.kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/perun/perun/collect/trace/run.py", line 99, in collect
    config.engine.assemble_collect_program(**kwargs)
  File "[...]/perun/perun/collect/trace/pin/engine.py", line 139, in assemble_collect_program
    self._assemble_pintool(**configuration)
  File "[...]/perun/perun/collect/trace/pin/engine.py", line 241, in _assemble_pintool
    loader=PackageLoader('perun', 'collect/trace/pin/templates'),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[...]/perun/.perun-venv-python3.11/lib64/python3.11/site-packages/jinja2/loaders.py", line 326, in __init__
    raise ValueError(
```
Note: The ``[...]`` is truncated path (e.g. ``/home/user/projects``).

The ``ValueError`` is triggered from the [jinja2/loaders.py](https://github.com/pallets/jinja/blob/9c6c3198996014756c111ac4d2e80b0423b3b95a/src/jinja2/loaders.py#L274) where one can see that 
[import spec](https://github.com/pallets/jinja/blob/9c6c3198996014756c111ac4d2e80b0423b3b95a/src/jinja2/loaders.py#L295C9-L295C54) is used to extract [submodule search locations](https://github.com/pallets/jinja/blob/9c6c3198996014756c111ac4d2e80b0423b3b95a/src/jinja2/loaders.py#L312). If we compare the value of ``submodule_search_locations`` in environment initialized with ``make install`` and another with ``make dev`` we see that the editable version from ``make dev`` environment contains path that is not existing.

Environment initialized with ``make dev`` reporting invalid path in ``submodule_search_locations``:
```
Python 3.11.9 (main, Apr 17 2024, 00:00:00) [GCC 13.2.1 20240316 (Red Hat 13.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib.util as u
>>> spec = u.find_spec("perun")
>>> spec.submodule_search_locations
['[...]/perun/.perun-venv-python3.11/lib64/python3.11/site-packages/_perun_toolsuite_editable_loader.py/perun']
``` 
Environment initialized with ``make install``:
```
Python 3.11.9 (main, Apr 17 2024, 00:00:00) [GCC 13.2.1 20240316 (Red Hat 13.2.1-7)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import importlib.util as u
>>> spec = u.find_spec("perun")
>>> spec.submodule_search_locations
['[...]/perun/perun']
``` 

This could be a problem with the build system since @JiriPavela pointed out that the behaviour with editable environment is not deterministic (however for me it happened every time so far).

tfiedor:


perun/view_diff/sankey/run.py
```diff
@@ -529,8 +530,7 @@ def generate_sankey_difference(lhs_profile: Profile, rhs_profile: Profile, **kwa
     lhs_header, rhs_header = diff_kit.generate_headers(lhs_profile, rhs_profile)
 
     # Note: we keep the autoescape=false, since we kindof believe we are not trying to fuck us up
```
JiriPavela: This seems like an obsolete comment?

JiriPavela (approved):
Looks good. I'm wondering whether we should perhaps add at least one tox environment to test editable installs as well, so that we can somewhat reliably discover such issues early.

tfiedor:
> Looks good. I'm wondering whether we should perhaps add at least one tox environment to test editable installs as well, so that we can somewhat reliably discover such issues early.

Good idea, done.

PR #237, opened 2024-07-17
Add perun import command

tfiedor:


perun/cli_groups/import_cli.py
```diff
@@ -0,0 +1,128 @@
...
+from perun.profile import imports
+from perun.utils.common import cli_kit
+
+
+@click.group("import")
+@click.option(
+    "--machine-info",
+    "-m",
+    type=click.Path(resolve_path=True, readable=True),
+    help="Imports machine info from file in JSON format (by default, machine info is loaded from the current hoot)."
```
JiriPavela: Seems like there is a typo: hoot -> host?

perun/cli_groups/import_cli.py
```diff
@@ -0,0 +1,128 @@
...
+    help="Sets [INT] samplings of the ith profiled command.",
+)
+@click.pass_context
+def perf_group(ctx: click.Context, **kwargs: Any) -> None:
+    """Imports Perun profiles from perf results
+
+    This supports either profiles collected in:
+
+      1. Binary format: e.g., `collected.data` files, that are results of `perf record`
+      2. Text format: result of `perf script` that parses the binary into user-friendly and parsing-frinedly text format
```
JiriPavela: Another typo: frinedly -> friendly.

perun/profile/imports.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    if machine_info is not None:
+        with open(machine_info, "r") as machine_handle:
+            return json.load(machine_handle)
+    else:
+        return environment.get_machine_specification()
+
+
+def get_param(cfg: dict[str, Any], param: str, index: int) -> Any:
+    """
```
JiriPavela: The docstring is missing a short description. The function seems rather simple and understandable on its own, but still, should we add at least a short description here?

utils/generate_machine_info.sh
```diff
@@ -0,0 +1,130 @@
...
+  echo -n -e "]"
+
+  echo -n "}"
+}
+
+if [[ -z "$output_file" ]]; then
+  get_machine_specification | sed 's/\\/\\\\/g'
+else
+  get_machine_specification | sed 's/\\/\\\\/g' > "$output_file"
+fi
```
JiriPavela: Seems like we're missing a newline here. Pre-commit is set up such that it checks only `*.py` files so it doesn't accidentally change something it shouldn't. Maybe we could add `*.sh` files as well?

JiriPavela (approved):
Great, adding support for importing external profiles seems like a good idea.

Maybe a quick idea for a slight improvement: would it make sense to allow the user to import the profiling configuration (i.e., cmd, workload, warmups, ...) using a single file, similarly to how the machine info is imported? I can see it being a bit mundane having to specify numerous parameters in the future if / when we decide to support more external profiles, e.g., Valgrind.

utils/generate_machine_info.sh
tfiedor: Sure, I will let you add this to precommit hook  in separate PR. I will fix the missing new line though.

tfiedor:
> Great, adding support for importing external profiles seems like a good idea.
> 
> Maybe a quick idea for a slight improvement: would it make sense to allow the user to import the profiling configuration (i.e., cmd, workload, warmups, ...) using a single file, similarly to how the machine info is imported? I can see it being a bit mundane having to specify numerous parameters in the future if / when we decide to support more external profiles, e.g., Valgrind.

This is good idea, but I would probably make this more generic and solve it in different PR. For now it is like this after consultation with JH, who was not a fan of the additional file :)

PR #256, opened 2024-08-16
Add support for importing data from ELK

tfiedor:
This PR adds support of reading profiles from ELK, which saves the queries in form of `{'queries': [{...}...]}"`.

Also adds support for `metadata` key.


perun/profile/imports.py
```diff
@@ -40,6 +41,15 @@ class ImportProfileSpec:
 
 
 class ImportedProfiles:
+    """
+    Note: I would reconsider this class or refactor it, removing the logical elements, it obfuscates the logic a little
+      and makes the functions less readable (there are not streams/pipes as is most of the logic/perun); I for one am
+      rather "fan" of generic functions that takes structures and returns structure than classes with methods/logic.
+    TODO: the import-dir could be removed by extracting this functionality to command-line callback and massage
+    the paths during the CLI parsing; hence assuming that the paths are correct when importing. I think the parameter
+    only complicates the code.
```
JiriPavela, line 50: Just a note: this wouldn't really work, as the `import-dir` path prefix is also applied to paths that can be found in CSV file(s). Although we could technically parse the files in the callback, it doesn't seem like the proper place to parse files with a bit more complicated format and error checking.

JiriPavela (approved):

tfiedor:
I will merge this, even though I will rework the import to the ELK (not happy with the flattening), but I wish the metadata to be merged already.

perun/profile/imports.py
tfiedor: Would extracting this to config hierarchy work? Have `lookup.path`, e.g. that would be gathered recursively through runtime, local and shared config and helper function that would resolve the path wrt to this set path? Then `--import-dir` would simply add to the `runtime()`, one could add it to local or shared to not have to use `import-dir` at all.

PR #257, opened 2024-09-24
Polish perun import and viewdiff

### LINKED ISSUE #229: Improve the visualizations of differences
Based on the feedback from the Lunch And Learn, we should extend our showdiff with the following:

- [x] Scale the rectangles based wrt biggest profile
- [x] #242
- [ ] Add collapsing section with basic metrics (top metrics, number of functions, number of levels, maximal traces, number of unique traces, etc.)
- [x] Add exit code to the display
- [x] Add `?` to tooltips, so it is clear there is something helpful
- [x] Add synchronized progress in flamegraphs (however, is this even possible???)

JiriPavela:
This PR addresses several issues with the current perun import and viewdiff modules. Among other things:

- Stats and metadata specification, internal representation and visualization have been unified across different perf and ELK import and viewdiff commands.
- Perf profile stats may now be specified both on CLI and in CSV files.
- Stats may now be aggregated in different ways depending on the type of stats value(s), aggregation key and comparison operator.
- Aggregated stats are now displayed in a nested table in viewdiff.
- Metadata may now be specified directly on CLI using strings or JSON files.
- Collapsing or showing different parts of profile headers in viewdiff now collapses or hides both LHS and RHS.

I think this issue resolves the last checkbox of #229 (please confirm @tfiedor), now that stats and metadata may be fully user-defined, including various aggregation and comparison functions, etc.


perun/cli_groups/import_cli.py
```diff
@@ -106,30 +123,83 @@ def perf_group(ctx: click.Context, **kwargs: Any) -> None:
...
 @click.pass_context
-def from_stacks(ctx: click.Context, imported: list[str], **kwargs: Any) -> None:
+def from_stacks(ctx: click.Context, import_entries: list[str], **kwargs: Any) -> None:
     """Import Perun profiles from output generated by `perf script | stackcollapse-perf.pl`
-    command
+    command.
+
+    Multiple import entries may be specified; an import entry is either a profile entry
+
+      'profile_path[,<exit code>[,<stat value>]+]'
```
tfiedor, line 186: Why this change? And can you maybe write an example how his is run? That's better readable then nested parenthesized specification.

tfiedor (approved):

perun/cli_groups/import_cli.py
JiriPavela: The parentheses represent optional parts of the import entries. This change was done so that (a) CSV and CLI specification of profiles to import is unified, i.e., it is now possible to specify exactly the same import parameters using both CSV files and CLI, and (b) at least exit code may be specified on CLI alongside the profiles to import. However, the interface is backwards-compatible and simple profile paths, e.g., `import.stack.gz` may still be specified. An example input containing both exit code and some stats values (corresponding to some specified stats headers defined in option `--stats-headers`) is, e.g., `'import.stack.gz,0,18511.379883,367'`.

PR #264, opened 2024-11-03
Add flamegraph diffs to the showdiff report

JiriPavela:
Originally, the showdiff report had some code to generate flamegraph diff, but the call to the generate_flamegraphs function had skip_diff=True, so no diff would be generated.

This PR re-enables flamegraph generation and moreover:

- Both baseline-target and target-baseline diffs are generated separately and shown directly below baseline or target flamegraph, respectively. The motivation behind generating both comparisons is discussed [here](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) (Section "Negation").
- All four flamegraphs are now unified: they have the same generation parameters where it makes sense (except minor differences, e.g., the --negation flag), they all respond to global search, have the same height and width, etc. Moreover, clicking into the diff flamegraph now opens the sankey graph as well.
- The height resizing of SVG has been reworked. Previously, the height of the images was set according to the longest known traces, however, flamegraph.pl does some postprocessing and removes subtraces that don't meet certain width threshold. This caused some of the generated SVGs to have too much unnecessary blank space above the graphs. E.g., in one case, the longest trace was 34 while the postprocessed one was only 16, which caused the image to have more than double the necessary height and made it harder to compare all four graphs at the same time.

### DIFF
--- a/perun/view/flamegraph/flamegraph.py
+++ b/perun/view/flamegraph/flamegraph.py
@@ -88,7 +88,6 @@ def draw_flame_graph(
     :param fg_max_resource: maximum number of samples collected
     """
     # converting profile format to format suitable to Flame graph visualization
-    # flame = convert.to_flame_graph_format(profile, profile_key=profile_key, minimize=minimize)
     with tempfile.NamedTemporaryFile(delete=False) as tmp:
         tmp.write("".join(flame_data).encode("utf-8"))
         tmp.close()
--- a/perun/view/flamegraph/flamegraph.py
+++ b/perun/view/flamegraph/flamegraph.py
@@ -3,64 +3,61 @@
 from __future__ import annotations
 
 # Standard Imports
-from typing import TYPE_CHECKING
+from dataclasses import dataclass, field
 import os
 import tempfile
+from typing import Any
 
 # Third-Party Imports
 
 # Perun Imports
-from perun.profile import convert
 from perun.utils import mapping
 from perun.utils.common import script_kit
 from perun.utils.external import commands
 
-if TYPE_CHECKING:
-    from perun.profile.factory import Profile
-
 
 def draw_flame_graph_difference(
-    lhs_profile: Profile,
-    rhs_profile: Profile,
-    width: int = 1200,
-    title: str = "",
-    profile_key: str = "amount",
-    minimize: bool = False,
+    lhs_flame_data: list[str],
+    rhs_flame_data: list[str],
+    title: str,
+    units: str = "samples",
+    img_width: int = 1200,
+    fg_flags: str = "",
+    fg_min_width: str = "1",
+    fg_max_trace: int = 0,
+    fg_max_resource: float = 0.0,
 ) -> str:
     """Draws difference of two flame graphs from two profiles
 
-    :param lhs_profile: baseline profile
-    :param rhs_profile: target_profile
-    :param width: width of the graph
-    :param profile_key: key for which we are constructing profile
-    :param title: if set to empty, then title will be generated
-    """
-    # First we create two flamegraph formats
-    lhs_flame = convert.to_flame_graph_format(
-        lhs_profile, profile_key=profile_key, minimize=minimize
-    )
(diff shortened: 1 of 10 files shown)


perun/view/flamegraph/flamegraph.py
```diff
@@ -69,35 +66,31 @@ def draw_flame_graph_difference(
...
     # converting profile format to format suitable to Flame graph visualization
-    flame = convert.to_flame_graph_format(profile, profile_key=profile_key, minimize=minimize)
-
-    header = profile["header"]
-    profile_type = header["type"]
-    cmd, workload = (header["cmd"], header["workload"])
-    title = title if title != "" else f"{profile_type} consumption of {cmd} {workload}"
-    units = mapping.get_unit(mapping.get_readable_key(profile_key))
-
+    # flame = convert.to_flame_graph_format(profile, profile_key=profile_key, minimize=minimize)
```
tfiedor: Why is this commented out here?

tfiedor (approved):
Can you elaborate in more details, what and why is changed here? It's not that clear, since flamegraph diffs were there initially as well.

perun/view/flamegraph/flamegraph.py
JiriPavela: Good catch, this comment shouldn't be there. I'll remove it.

JiriPavela:
> Can you elaborate in more details, what and why is changed here? It's not that clear, since flamegraph diffs were there initially as well.

Sure, originally, the showdiff report had some code to generate flamegraph diff, but the call to the `generate_flamegraphs` function had `skip_diff=True`, so no diff would be generated.

This PR re-enables flamegraph generation and moreover:
- Both baseline-target and target-baseline diffs are generated separately and shown directly below baseline or target flamegraph, respectively. The motivation behind generating both comparisons is discussed [here](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) (Section "Negation").
- All four flamegraphs are now unified: they have the same generation parameters where it makes sense (except minor differences, e.g., the `--negation` flag), they all respond to global search, have the same height and width, etc. Moreover, clicking into the diff flamegraph now opens the sankey graph as well.
- The height resizing of SVG has been reworked. Previously, the height of the images was set according to the longest known traces, however, `flamegraph.pl` does some postprocessing and removes subtraces that don't meet certain width threshold. This caused some of the generated SVGs to have too much unnecessary blank space above the graphs. E.g., in one case, the longest trace was 34 while the postprocessed one was only 16, which caused the image to have more than double the necessary height and made it harder to compare all four graphs at the same time.

tfiedor:
> > Can you elaborate in more details, what and why is changed here? It's not that clear, since flamegraph diffs were there initially as well.
> 
> Sure, originally, the showdiff report had some code to generate flamegraph diff, but the call to the `generate_flamegraphs` function had `skip_diff=True`, so no diff would be generated.
> 
> This PR re-enables flamegraph generation and moreover:
> 
> * Both baseline-target and target-baseline diffs are generated separately and shown directly below baseline or target flamegraph, respectively. The motivation behind generating both comparisons is discussed [here](https://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) (Section "Negation").
> * All four flamegraphs are now unified: they have the same generation parameters where it makes sense (except minor differences, e.g., the `--negation` flag), they all respond to global search, have the same height and width, etc. Moreover, clicking into the diff flamegraph now opens the sankey graph as well.
> * The height resizing of SVG has been reworked. Previously, the height of the images was set according to the longest known traces, however, `flamegraph.pl` does some postprocessing and removes subtraces that don't meet certain width threshold. This caused some of the generated SVGs to have too much unnecessary blank space above the graphs. E.g., in one case, the longest trace was 34 while the postprocessed one was only 16, which caused the image to have more than double the necessary height and made it harder to compare all four graphs at the same time.

Nice, thank, sounds good. Please copy this to the PR description. Very good for fixing the height, it was quite hacked.

PR #265, opened 2024-11-03
Add CPU vulnerabilities to showdiff

### LINKED ISSUE #263: Display info about Vulnerabilities in Profile Specification
Hi all,

could we add info about Vulnerabilities in Profile Specification?

Vulnerabilities are a common source of performance changes. Perun already has the information from lscpu:

```
lscpu
....
Vulnerabilities:          
  Gather data sampling:   Mitigation; Microcode
  Itlb multihit:          Not affected
  L1tf:                   Not affected
  Mds:                    Not affected
  Meltdown:               Not affected
  Mmio stale data:        Not affected
  Reg file data sampling: Not affected
  Retbleed:               Not affected
  Spec rstack overflow:   Not affected
  Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
  Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; RSB filling; PBRSB-eIBRS SW sequence; BHI SW loop, KVM SW loop
  Srbds:                  Not affected
  Tsx async abort:        Not affected
```
Please display information about vulnerabilities in profile specifications. This can be collapsed by default and marked red to notify users to expand it and check it. 

Thanks a lot
Jirka

JiriPavela:
This PR adds support for displaying CPU vulnerabilities in the showdiff commands.

CPU vulnerabilities are obtained using either the `lscpu` command or from `/sys/devices/system/cpu/vulnerabilities/`. Since CPU vulnerabilities may affect kernel performance, it is important to display them in the profile diffs. Moreover, this PR refactors and simplifies the code reponsible for generating profile headers and metadata in the showdiff package.

This PR resolves #263.

### DIFF
--- a/perun/cli.py
+++ b/perun/cli.py
@@ -64,6 +64,7 @@ from perun.utils.exceptions import (
     ExternalEditorErrorException,
 )
 from perun.utils.structs.common_structs import Executable
+from perun.utils.structs.diff_structs import HeaderDisplayStyle
 from perun import fuzz as fuzz
 import perun.postprocess
 import perun.profile.helpers as profiles
@@ -703,6 +704,16 @@ def show(ctx: click.Context, profile: Profile, **_: Any) -> None:
     is_flag=True,
     default=False,
     help="Creates self-contained outputs usable in offline environments (default=False).",
+)
+@click.option(
+    "--display-style",
+    "-d",
+    type=click.Choice(HeaderDisplayStyle.supported()),
+    default=HeaderDisplayStyle.default(),
+    callback=cli_kit.set_config_option_from_flag(perun_config.runtime, "showdiff.display_style"),
+    help="Selects the display style of profile header. The 'full' option displays all provided "
+    "headers, while the 'diff' option shows only headers with different values "
+    f"(default={HeaderDisplayStyle.default()}).",
 )
 @click.pass_context
 def showdiff(ctx: click.Context, **kwargs: Any) -> None:
--- a/perun/profile/helpers.py
+++ b/perun/profile/helpers.py
@@ -23,7 +23,7 @@ import operator
 import os
 import re
 import time
-from typing import Any, TYPE_CHECKING
+from typing import Any, TYPE_CHECKING, Union
 
 # Third-Party Imports
 
@@ -43,6 +43,9 @@ from perun.vcs import vcs_kit
 
 if TYPE_CHECKING:
     import types
+
+# A tuple representation of the ProfileHeaderEntry object
+ProfileHeaderTuple = tuple[str, Union[str, float], str, dict[str, Union[str, float]]]
 
 
 PROFILE_COUNTER: int = 0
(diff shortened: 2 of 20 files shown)


perun/cli.py
```diff
@@ -739,7 +750,7 @@ def showdiff(ctx: click.Context, **kwargs: Any) -> None:
 
         perun showdiff -m HEAD~1 0@i 1@i report
     """
-    pass
+    print("Here")
```
tfiedor: Destroy this.

perun/profile/helpers.py
```diff
@@ -44,6 +44,9 @@
 if TYPE_CHECKING:
     import types
 
+# A tuple representation of the ProfileHeaderEntry object
+ProfileHeaderTuple = tuple[str, Union[str, float], str, dict[str, Union[str, float]]]
```
tfiedor, line 48: Should we maybe start destroying this tuples and use dataclasses instead?

perun/profile/helpers.py
```diff
@@ -613,41 +616,64 @@ def is_compatible_with_profile(self, profile: profiles.Profile) -> bool:
...
+
+    def as_tuple(self) -> ProfileHeaderTuple:
+        """Converts the header object into a tuple.
+
+        :return: the tuple representation of a header entry
+        """
+        return self.name, self.value, self.description, self.details
+
+    @staticmethod
+    def _try_convert_value(value: str) -> str | float:
```
tfiedor: There is `try_convert` value in utils, that does exactly this.

utils/generate_machine_info.sh
```diff
@@ -118,7 +118,23 @@ get_machine_specification() {
       fi
     done < /proc/cpuinfo
   fi
-  echo -n -e "]"
+  echo -n -e "],"
+
+  vulns=$(lscpu | grep '^Vulnerability' | sed 's/Vulnerability //g' | sort)
```
tfiedor: I suggest fixing this to `grep '^Vulnerability '`, so this is safely captured (in case somehow you get text with "VulnerabilityRecord".

tfiedor (approved):

perun/cli.py
JiriPavela: Fixed.

perun/profile/helpers.py
JiriPavela: We now actually mostly use dataclasses in the showdiff, however, this tuple type is used to propagate the results to jinja template, which is then easier to process in jinja.

perun/profile/helpers.py
JiriPavela: Great, replaced. I also changed the `try_convert` function to not use `SuppressedException` context manager, as it adds unnecessary overhead when called many times in a loop.

utils/generate_machine_info.sh
JiriPavela: Fixed.

PR #277, opened 2025-01-01
Improved profile sorting and fields in status

### LINKED ISSUE #271: Allow the users to specify the resulting profile name in the `import` command
Right now, users cannot specify the output profile name in the `import` command. Other profile-related commands, such as `collect` allow this, so we should add support for this feature to `import` as well.

### LINKED ISSUE #272: Rework `status` ordering of profiles
Perun `status` currently indexes profiles from the most to least recent (0@p, 1@p, ...), which makes it difficult to identify profiles using those indices in an automated manner, as the profile indices change. We should change the indexing from least to most recent, so the individual profiles retain their indices.

### LINKED ISSUE #273: Add support for string profile tags
Currently, distinguishing between profiles stored in Perun is possible either through their name or numeric tag. Both profile names and numeric indices have to be unique, so that profiles can be unambiguously identified. However, when manually working with different profiles, it might be useful to tag certain profiles with (not necessarily unique) short strings that convey some meaning to the users, and allows them to faster distinguish which profiles store what data.

Such string tag should be stored within the profile and be displayed in `perun status`. Using the tag to identify profiles is, however, problematic as the string tag might not be unique. We should either add support for advanced indexing using string tags as well (e.g., 0@mytag), or not support such identification of profiles.

JiriPavela:
Perun config now supports option `format.sort_profiles_order` that allows to set the profile sort order as ascending or
descending. By default, profiles are now sorted in ascending order. This fixes #272.

Perun import now has '--profile-name' option to specify a custom profile name for the imported profile. This fixes #271.

Users can now add custom string labels that briefly describe profiles by some common characteristic, e.g., environment settings or machine specification. Labels can be currently specified only during `perun import`. This fixes #273.

### DIFF
--- a/perun/profile/helpers.py
+++ b/perun/profile/helpers.py
@@ -30,7 +30,7 @@ from typing import Any, TYPE_CHECKING, Union
 # Perun Imports
 from perun.logic import config, index, pcs, store
 from perun import profile as profiles
-from perun.utils import log as perun_log
+from perun.utils import decorators, log as perun_log
 from perun.utils.common import common_kit
 from perun.utils.external import environment
 from perun.utils.exceptions import (
--- a/perun/profile/helpers.py
+++ b/perun/profile/helpers.py
@@ -105,6 +105,8 @@ def generate_profile_name(profile: profiles.Profile) -> str:
             Command of the job
         `%workload%`:
             Workload of the job
+        `%label%`:
+            A custom string label associated with the profile
         `%type%`:
             Type of the generated profile
         `%kernel%`:
@@ -155,6 +157,10 @@ def generate_profile_name(profile: profiles.Profile) -> str:
                 )
                 + "]",
             ),
+            (
+                r"%label%",
+                lambda scanner, token: lookup_value(profile["header"], "label", "_"),
+            ),
             (r"%kernel%", lambda scanner, token: environment.get_kernel()),
             (
                 r"%type%",
@@ -201,6 +207,7 @@ def load_list_for_minor_version(minor_version: str) -> list["ProfileInfo"]:
                 "type": index_entry.type,
                 "cmd": index_entry.cmd,
                 "workload": index_entry.workload,
+                "label": index_entry.label,
             },
             "collector_info": {"name": index_entry.collector},
             "postprocessors": [{"name": p} for p in index_entry.postprocessors],
@@ -518,6 +525,7 @@ class ProfileInfo:
         "type",
         "cmd",
         "workload",
+        "label",
         "collector",
         "postprocessors",
         "checksum",
@@ -540,7 +548,6 @@ class ProfileInfo:
         :param is_raw_profile: true if the stored profile is raw, i.e. in json and not
             compressed
         """
-
         self._is_raw_profile = is_raw_profile
         self.source = path
         self.realpath = os.path.relpath(real_path, os.getcwd())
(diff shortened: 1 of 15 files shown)


perun/profile/helpers.py
```diff
@@ -408,47 +414,32 @@ def is_key_aggregatable_by(profile: profiles.Profile, func: str, key: str, keyna
     return True
 
 
-@decorators.singleton
```
tfiedor, line 411: This was actually kept as an optimization, mainly for test cases, since it is called a lot, and the recursive lookup of keys in the hierarchy is costly. Since you removed this function completely, it would be better to move the caching somewhere. However, I'm currently not keen on moving it to `safely_lookup_recursively`. I'd rather have this function `get_sort_order`, which would call the two lookups instead.

perun/utils/structs/common_structs.py
```diff
@@ -677,6 +677,36 @@ def __exit__(self, exc_type: str, exc_val: Exception, exc_tb: traceback.StackSum
         return isinstance(exc_val, self.handler_exc)
 
 
+class SortOrder(Enum):
+    """Enumeration representing sorting order in a more descriptive way compared to the library."""
+
+    Ascending = "ascending"
+    Descending = "descending"
```
tfiedor: I seriously can't be bothered to write ascending/descending. I would suggest to shorten this to `asc` and `desc`; it's well-established.

tfiedor (approved):
What needs to be addressed before merge:
1. Check the runtime of the tests (for this PR and baseline): whether there is a degradation due to removing caching from getting sort order. Unless it is negligible, (fractions of seconds), fix the caching.

JiriPavela:
> What needs to be addressed before merge:
> 
>    1. Check the runtime of the tests (for this PR and baseline): whether there is a degradation due to removing caching from getting sort order. Unless it is negligible, (fractions of seconds), fix the caching.

Good point, I measured both the entire test suite and just the `perun status` tests, and checked the tests runtime in our CI as well to see if, and how much, the caching helps.

- Total runtime of tests ran locally increased by about 0.9s (82.949s vs. 82.052s).
- Runtime of status tests ran locally increased by about 0.01s (3.038s vs. 3.028s).
- CI tests run either < 1s slower, or in case of py3.12 and 3.13 faster by up to 7 seconds.

Overall, the omission of caching seems to have no statistically significant impact on the runtime of tests, however, I'll add the caching regardless.

tfiedor:
> > What needs to be addressed before merge:
> > 
> > 1. Check the runtime of the tests (for this PR and baseline): whether there is a degradation due to removing caching from getting sort order. Unless it is negligible, (fractions of seconds), fix the caching.
> 
> Good point, I measured both the entire test suite and just the `perun status` tests, and checked the tests runtime in our CI as well to see if, and how much, the caching helps.
> 
> * Total runtime of tests ran locally increased by about 0.9s (82.949s vs. 82.052s).
> * Runtime of status tests ran locally increased by about 0.01s (3.038s vs. 3.028s).
> * CI tests run either < 1s slower, or in case of py3.12 and 3.13 faster by up to 7 seconds.
> 
> Overall, the omission of caching seems to have no statistically significant impact on the runtime of tests, however, I'll add the caching regardless.

Naturally, this is not a considerable regression for user experience (this is basically called only once during a perun command, hence there is no real difference for running `perun status`), however, for tests this is called repeatedly, hence, the ~1s regression, which might be "insignificant" from statistical point of view, however, its needless to not cache it (and I managed to reduce the test harness duration thanks to such minor optimizations, so I'd rather keep the test case duration like that).

PR #290, opened 2025-03-03
Feature: forward parameters to flamegraph

JiriPavela:
This PR extends the CLI of `showdiff report` so that the user can forward a curated subset of parameters to the `flamegraph.pl` script.

### DIFF
--- a/perun/collect/kperf/parser.py
+++ b/perun/collect/kperf/parser.py
@@ -28,7 +28,7 @@ def parse_events(perf_events: list[str]) -> list[dict[str, Any]]:
         if event.strip():
             *record, samples = event.split(" ")
             parts = " ".join(record).split(";")
-            command, trace, uid = parts[0], parts[1:-1], parts[-1]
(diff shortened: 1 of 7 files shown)


tfiedor (approved):
Seems OK. The question is, if it would be possible to forward the paramaters similar to what other projects do, with `--`, i.e. `perun show flamegraph -- -w 10 -h 10`, etc.

JiriPavela:
There seems to be no direct support for this in Click as far as I looked. It is possible to pass unknown arguments (https://click.palletsprojects.com/en/stable/advanced/#forwarding-unknown-options) but the syntax is different. Since we sometimes need to do some postprocessing on the supplied parameters (e.g., `width` and `minwidth`), it is easier to let Click parse them directly than to parse them from the unknown remaining arguments. Moreover, we do not want the users to be able to forward every `flamegraph.pl` parameter, as they might accidentally overwrite parameters that we pass to the script (e.g., `total` or multiple options related to color or consistent palette).

PR #294, opened 2025-03-19
Enhanced diff report 

robertvasek:
This pull request introduces new facelift of diff report.

*A showcase of the new report can be found in this attachment:*
[perun_mmap_offline.html.zip](https://github.com/user-attachments/files/19339650/perun_mmap_offline.html.zip)


JiriPavela (commented):
I reviewed the changes and I think we are ready for a merge. Waiting for a review from @tfiedor.

tfiedor (approved):
I have faith that Jirka has went through the templates ;), I mainly checked the python code and they seem fine to me.

I have two main suggestions:
1.  I think the "pin" icon is confusing to use to add some note (I get that it is from pin the note, but I think generally the icon is used to pin something so it is displayed and not hidden, so we could consider some different icon; or some modification, like pin with paper). However, what it definitely needs is a tooltip: "Add note".
2. I would consider creating second simplified design, with minimal shading and minimal shapes; something friendly for paper and paper-like.

JiriPavela:
Good points, the pin icon suggestions sound good to me. As for the paper version of the report – maybe supporting some form of print/pdf version might be a good way to do it.

JiriPavela (approved):

PR #298, opened 2025-03-27
Report facelift - issues fix

robertvasek:
New version of report facelift, which primarily fixes some issues.

1. Table traces are not creating `inception` when Exported with notes
2. Note icons changed to chat bubbles
3. Report is more condensed to that user doesn't have to scroll that much
4. Nested table inside table traces has fixed details
5. Sections are ordered differently
6. Embedded icons cannot be deleted

*New report version is also attached in .zip file.*
[perun_mmap_offline.html.zip](https://github.com/user-attachments/files/19485798/perun_mmap_offline.html.zip)


JiriPavela (approved):
Great, we're ready for a merge.

PR #301, opened 2025-04-27
Unify naming and saving of profiles

JiriPavela:
This PR refactors the naming and saving of Perun profiles, removing some existing code duplication.

- Profiles are no longer overwritten by default if an existing profile with identical name exists. Instead, the profile name is suffixed with a copy number, e.g., `custom_profile(N).perf`.
- The collect and import CLI now define identical options `--profile-name` and `--profile-dir` that determine the name and target directory for storage of the profile.
- Profiles may now be sorted w.r.t. multiple keys and orderings.
- New sort keys `stem` and `copy` were added to handle new naming scheme with name copies.
- By default, the sorting is now done using multikey `time` > `stem` > `copy` that results in a more natural ordering of profiles.

### DIFF
--- a/perun/cli.py
+++ b/perun/cli.py
@@ -42,9 +42,10 @@ the flexibility of Perun's usage.
 from __future__ import annotations
 
 # Standard Imports
-from typing import Optional, Any, TYPE_CHECKING
+import functools
 import os
 import sys
+from typing import Optional, Any, TYPE_CHECKING
 
 # Third-Party Imports
 import click
@@ -562,10 +563,17 @@ def log(head: Optional[str], **kwargs: Any) -> None:
     "-sb",
     "format__sort_profiles_by",
     nargs=1,
-    type=click.Choice(profiles.ProfileInfo.valid_attributes),
-    callback=cli_kit.set_config_option_from_flag(pcs.local_config, "format.sort_profiles_by", str),
+    type=str,
+    callback=cli_kit.set_config_option_from_flag(
+        pcs.local_config,
+        "format.sort_profiles_by",
+        functools.partial(
+            cli_kit.config_value_parse_and_validate,
+            allowed_values=profiles.ProfileInfo.valid_attributes,
+        ),
+    ),
     help=(
-        "Sets the sort key in the local configuration as 'format.sort_profiles_by' that will be "
+        "Sets the sort keys in the local configuration as 'format.sort_profiles_by' that will be "
         "used for sorting both pending and index profiles."
     ),
 )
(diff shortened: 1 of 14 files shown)


perun/cli.py
```diff
@@ -562,10 +563,17 @@ def log(head: Optional[str], **kwargs: Any) -> None:
     "-sb",
     "format__sort_profiles_by",
     nargs=1,
-    type=click.Choice(profiles.ProfileInfo.valid_attributes),
-    callback=cli_kit.set_config_option_from_flag(pcs.local_config, "format.sort_profiles_by", str),
+    type=str,
+    callback=cli_kit.set_config_option_from_flag(
+        pcs.local_config,
+        "format.sort_profiles_by",
+        functools.partial(
```
tfiedor: This is really ugly, and IMO should not be in CLI at all, if you need it like this, then extract it to some function so this partial is hidden.

perun/logic/runner.py
```diff
@@ -473,7 +437,9 @@ def run_postprocessor_on_profile(
 
     p_status, processed_profile = run_postprocessor(postprocessor_unit, profile_job, prof)
     if p_status == PostprocessStatus.OK and prof and not skip_store:
-        store_generated_profile(processed_profile, profile_job)
+        profile.save_profile(
+            profile.finalize_profile_for_job(prof, profile_job, prof.get("header", {}).get("label"))
```
tfiedor, line 441: This will fail if there is no header.

tfiedor (approved):
Seems fine. Just please extract that partial to some function/factory.

perun/cli.py
JiriPavela: Good point, perhaps the `config.py` module could keep track of allowed values (or other constraints, if any) for individual configuration keys and implement the validity checks instead of cli_kit callbacks?

perun/logic/runner.py
JiriPavela: Could you please elaborate this a bit more? The `prof.get("header", {})` is there exactly for situations when a header is not present in a profile, in which case it will use the default value of empty dictionary and as a result, the label parameter will be set to `None`.

tfiedor: Good point, for some reason my brain though that `{}.get("label")` would end with error. It is fine then.

perun/cli.py
JiriPavela: I replaced the `partial` with its alias so that it is more readable. I also created an issue #302 related to the problem of systematic checking config keys and values.

tfiedor: Mm, yes, I guess ideal would be to have some json/yaml with the exact specification that would check the validity of the config (together with types, expected ranges, etc.).

PR #304, opened 2025-05-07
REPORT: Note's modal window popup

robertvasek:
This pull request introduces new modal window for notes in reports.

- When the user clicks on the `chat bubble` icon, a popup appears, allowing them to edit, save, or delete the note.
- Additionally, an option to mark the note as `Attention` or `Waived` is available below.
- When the user has unsaved changes, the `Save` button is light red (or uses the `common` color from the profile color palette); when saved, the color is purple (or the `target` color from the profile color palette).
- The note section now merely serves as a summary of created notes, and when the user clicks on a note title in the note section, the modal window appears again.
- All of the above has resulted in the notes exporting functionality being heavily simplified.

_As an illustration, sample file with new enhancements is attached._
([perun_mmap_offline.html.zip](https://github.com/user-attachments/files/20084006/perun_mmap_offline.html.zip))


JiriPavela (approved):
Great, I like the new modal window for managing notes. I have three suggestions:

1. I would consider adding another button into the modal window, named for example `Save and close`. Sometimes the users might want to just save their progress and continue typing, but they might also want to finish the note and continue. Currently, the `Save` button works a bit unintuitively in the second case.
2. I think it might be helpful to add some "jump to" icon to the individual notes in the sidebar. Clicking on the note will jump to the location but also open the modal window, which should probably be two separate actions, i.e., jump or edit.
3. Would it make also sense to implement some periodic auto-save when writing notes? I can imagine scenarios, where performance engineers are writing a long explanation and then accidentally lose it all.

robertvasek:
The auto-save depends. The idea is that person sometimes writes something and in the process realizes that it actually does not make sense and wants to erase it all. Or not?

JiriPavela:
> The auto-save depends. The idea is that person sometimes writes something and in the process realizes that it actually does not make sense and wants to erase it all. Or not?

Sure, in that case, there should still be some way to discard the note. I am merely concerned that without some form of autosave, the engineers might write a few paragraphs of description and then accidentally close or refresh the page, at which point the draft will likely be lost. But I will first try to play with it a bit to see if and when the notes might get lost, and if we thus really need or want some form of autosave.

JiriPavela:
Other than that, I think we are ready for a merge. Good job.

Issue #305, opened 2025-05-07
Inefficient processing of profiling data in `perun showdiff report`


JiriPavela:
Perun does multiple passes through the profiling data when building the report. This introduces significant inefficiency, namely for large profiles.

One of the solution is to implement a suitable data structure for storing and postprocessing the profiling data from `perf` or other profilers and tracers. [Calling-context Tree (CCT)](https://dl.acm.org/doi/abs/10.1145/258916.258924) is one such data structure most suited for context-sensitive performance data. An efficient C++ implementation with Python bindings is currently being developed, however, a simple prototype Python implementation for speeding up Perun report construction should be developed as well.

PR #306, opened 2025-05-07
Improve handling of tall flamegraphs

JiriPavela:
Perun report currently does not handle tall flamegraphs in a user-friendly manner. This PR attempts to improve this in two ways.
- Flamegraphs now by-default squash `[unknown]` frames into a single frame to avoid tall `[unknown]` towers. This feature may be disabled by the CLI option `--no-squash-unknown`. Note that this is more of an ad-hoc fix, as a proper systematic solution needs #305 to be resolved first in order to avoid computationally expensive processing of the profiling data.
- Additionally, report now supports forwarding the `--inverted` flag to the `flamegraph.pl` script, which creates icicle graphs instead of flame graphs. This is useful for large profiles that do not fully fit into the viewport.

### DIFF
--- a/perun/profile/convert.py
+++ b/perun/profile/convert.py
@@ -107,7 +107,10 @@ def models_to_pandas_dataframe(profile: Profile) -> pandas.DataFrame:
 
 
 def to_flame_graph_format(
-    profile: Profile, profile_key: str = "amount", minimize: bool = False
+    profile: Profile,
+    profile_key: str = "amount",
+    minimize: bool = False,
+    squash_unknown: bool = True,
 ) -> list[str]:
     """Transforms the **memory** profile w.r.t. :ref:`profile-spec` into the
     format supported by perl script of Brendan Gregg.
@@ -135,6 +138,7 @@ def to_flame_graph_format(
     :param profile: the memory profile
     :param profile_key: key that is used to obtain the data
     :param minimize: minimizes the uids
+    :param squash_unknown: whether recursive [unknown] frames should be squashed into a single one
     :returns: list of lines, each representing one allocation call stack
     """
     stacks = []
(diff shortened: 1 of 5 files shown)


perun/view/flamegraph/flamegraph.py
```diff
@@ -145,7 +145,11 @@ def build_flamegraph_command(
     cmd.extend(f"--{flag}" for flag in flags)
     # Additional parameters
     for key, val in kwargs.items():
-        if val is not None:
+        if val is True:
+            # This kwarg is a flag
+            cmd.append(f"--{key}")
+        elif val not in (None, False):
```
tfiedor: Should there be else clause?

perun/view_diff/report/run.py
```diff
@@ -740,6 +740,7 @@ def generate_report(lhs_profile: Profile, rhs_profile: Profile, **kwargs: Any) -
         Stats.all_stats(),
         skip_diff=False,
         minimize=Config().minimize,
+        squash_unknown=not kwargs.get("no_squash_unknown", False),
```
tfiedor, line 743: Are we sure about this? This double negative could be not easy to read IMO (see what I did there :D).

perun/profile/convert.py
```diff
@@ -143,15 +147,27 @@ def to_flame_graph_format(
...
+                                unknown_cnt += 1
+                                continue
+                            elif unknown_cnt > 0:
+                                stack_str += "[unknown[squashed]];"
+                                unknown_cnt = 0
+                        stack_str += f"{line};"
+                    # The unknown frame might have been the last one, in which case the frame was
+                    # not yet added to the string.
+                    if unknown_cnt > 0:
+                        stack_str += "[unknown[squashed]];"
```
tfiedor, line 168: Would it make sense to output this as `[unknown[squashed]]^{unknown_cnt}` or something similar to know how much you squashed?

tfiedor (approved):

perun/view/flamegraph/flamegraph.py
JiriPavela: Good point, it should actually be `elif val`, as simple `else` would execute even for `val`s set to None.

perun/view_diff/report/run.py
JiriPavela: A reasonable alternative would perhaps be changing the bool function parameter to `no_squash_unknown` to match the CLI flag. The CLI flag should stay negative, as the squashing is by-default enabled.

perun/profile/convert.py
JiriPavela: Ideally, yes, that was the original intention. However, it (a) may break a single parent unknown frame into multiple smaller frames, which changes the graph topology in a bit unintuitive way, and (b) it complicates the mouseover highlighting, as it relies on exact match. As stated in the PR message, this is more of an ad-hoc fix that I plan to redo later as part of refactoring the data handling in report.

perun/view_diff/report/run.py
JiriPavela: Looking at the issue a bit closer, it will actually make it even worse in the code that uses the flag. There does not seem to be a good solution for negative flags, but the current version at least limits it to a single line of code.

perun/profile/convert.py
tfiedor: I thought it squashed all unknows in tower to single unit. Can you maybe show a picture/examples how this squashes the towers?

JiriPavela: The current solution does exactly that, but the proposed solution with `[unknown[squashed]]^{unknown_cnt}` would not, as the towers are in fact often trees (i.e., they branch into multiple branches all having a different leaf function). Hence, the tree would be split into strictly linear traces, all with different names (e.g., `[unknown[squashed]]^2`, `[unknown[squashed]]^5`, etc.) and the flame graph algorithm would not be able to merge them that way.

PR #315, opened 2025-07-02
Refactor the view_diff package

### LINKED ISSUE #312: Update the flamegraph difference view to support features from the diff report
The goal is to port the features that were implemented in the novel diff report to flamegraph diff view (the grid, the global search, etc.).

JiriPavela:
This PR removes obsolete `view_diff` modules (`sankey` and `datatables`) and flattens the package structure so that the remaining diffs are now modules and not packages.

The `flamegraph` diff has been updated to use the same page structure as the new `report` diff. However, compared to the `report`, it renders only the flame graph grid and statistics. This resolves the #312 issue.

The PR also decouples the `view_diff` CLI and the logic of the diffs, and implements lazy loading of all `view_diff` modules.


perun/utils/structs/diff_structs.py
```diff
@@ -29,3 +38,44 @@ def default() -> str:
         :return: the default display style
         """
         return HeaderDisplayStyle.FULL.value
+
+
+def singleton_class(cls: Type[Any]) -> Callable[[], Config]:
```
tfiedor, line 43: Can this be extracted e.g. to decorators or structs?

tfiedor (approved):

perun/utils/structs/diff_structs.py
JiriPavela: Ideally yes, however, I plan to refactor the global singleton instance out when I update the diffs memory model (which is already under development). Until then, I just moved the existing `Config` class to the `diff_structs` file in order to make lazy loading possible.

tfiedor: Then add some TODO there so it is refactored later, maybe with issue number if you have one open for that.

Issue #314, opened 2025-06-10
Allow to customize the layout of the diff reports based on configuration


tfiedor:
The current report is heavily specific for certain usage of certain users. We need to open this to more broader audience: one of the things that could help is to allow users to pick a parts of the report that would be displayed + the order. This could be specified in configuration e.g. similar to some GUI applications that use YAML to specify the layout.

E.g. one could want only the flamegraphs and the table, no summary, no vulnerabilities.

PR #316, opened 2025-07-14
Cleanup `perun/templates` directory.

JiriPavela:
This PR refactors and cleans up the `templates` directory within Perun. The templates for new checks, collectors, etc. are now in a separate directory `units` and the diff view templates are in the `diff_views` directory. Moreover, the diff views templates have been modularized so that it is easy to create new views out of the existing sections and components. This is a prerequisite work for #314.


tfiedor (approved):
LGTM, nice job.

PR #319, opened 2025-07-19
Extend the showdiff traces table with new metrics

JiriPavela:
The traces table in reports now contain a new set of metrics: baseline (absolute), target (absolute), total delta, absolute delta and relative delta. The table now also contains tooltips on certain columns, and some minor styling changes have been made. The most important new metric is the `total delta` which shows the total impact of the change on the overall number of consumed resources.

The extended table looks like this:
Screenshot from 2025-07-19 22-10-14

### DIFF
--- a/perun/templates/diff_sections/section_traces_table.html.jinja2
+++ b/perun/templates/diff_sections/section_traces_table.html.jinja2
@@ -1,6 +1,7 @@
 {% import 'macros/section_title.html.jinja2' as m_title %}
 
-{% set traces_table_info = "Click on a row to explore top N calling contexts (traces) of the selected function.
+{% set traces_table_info = "Mouseover a column header to display a tooltip explaining the column.
+            Click on a row to explore top N calling contexts (traces) of the selected function.
             Click on (+) next to a trace row to open a detail about that concrete trace.
             You may filter the table using the select form in the table header." %}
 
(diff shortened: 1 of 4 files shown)


perun/templates/diff_sections/section_traces_table.html.jinja2
```diff
@@ -1,22 +1,26 @@
...
+    <table id="table" class="traces_table display" style="width: 100%;">
         <thead>
             <tr>
                 <th></th>
                 <th>Unit</th>
                 <th>Change</th>
                 <th>Metric</th>
-                <th>Absolute Δ</th>
-                <th>Relative Δ</th>
+                <th title="The amount of resources consumed by the Unit in the baseline profile.">Baseline</th>
```
tfiedor, line 19: Looking at this get me thinking: should we extract this (I mean the columns) to some constant/variable in the code? That could be more maintenable if we decide to do this differently.

perun/templates/scripts/traces_table_js.html.jinja2
```diff
@@ -131,7 +153,7 @@
                 paging: false,
                 info: false,
                 searching: false,
-                order: [[3, "desc"]]
+                order: [[5, "desc"]]
```
tfiedor, line 156: Not sure how constant-friendly is the JS here, but maybe extracting these columns to variables would be safer.

perun/view_diff/report.py
```diff
@@ -567,6 +588,33 @@ def generate_trace_stats(graph: Graph) -> dict[str, list[TraceStat]]:
...
+        for stat in Config().profile_stats["baseline"]
+        if stat.name.startswith("Overall")
+    ], [
+        float(stat.value[0])
+        for stat in Config().profile_stats["target"]
+        if stat.name.startswith("Overall")
+    ]
+
+
+def _make_percentage(value: float) -> float:
```
tfiedor: `_to_percentage`?  `_as_percentage`? That would sound more clearer, `make` sounds very vague.

tfiedor (approved):
LGTM; I suggested few refactoring for future; there is lot of magic numbers (I marked few) that would be nice to extract.

JiriPavela:
Thanks for the suggestion, I agree the HTML/JS part of the traces table should be refactored. Currently, the table structure and data providers are hard-coded at multiple code locations (usually with some minor adjustments), so it would require much more effort to make the table properly maintainable and easily customizable in terms of changing columns and/or their order. I created a separate issue for this (#320) and will leave it to a standalone future PR.

PR #325, opened 2025-09-10
REPORT: NAVIGATION PANEL LINKS

robertvasek:
This PR allows user to add embedded links into report. Usage is as follows:
`perun showdiff --offline 28@p 68@p report -o report.html --link https://muni.cz MUNI`

The links are then displayed in the navigation panel in the last place.
User can add more of these links or none, although two links are pernamently embedded:
- Perun repository
- Brendan Gregg Flame Graph Repository

Details can be found in the attachments, where exemplary report is placed.
[report_x.html.zip](https://github.com/user-attachments/files/22258850/report_x.html.zip)

Also 3 fuzzing tests are skipped on MacOS. These tests are successful locally but are failing because of the incorrect pipeline setup (clang compiler is used instead of gcc). This issue is to be addressed later by me and @JiriPavela.


perun/templates/macros/traces_selection.html.jinja2
```diff
@@ -0,0 +1,77 @@
+{% macro select_top_x_traces1(x, name, data) %}
```
JiriPavela, line 1: These macros are currently not being used in the resulting report, right? I just want to make sure we do not accidentally push mock data into production.

tests/test_fuzzing.py
```diff
@@ -23,6 +23,7 @@
 
 
 @pytest.mark.usefixtures("cleandir")
+@pytest.mark.skipif(sys.platform == "darwin", reason="Import perf record is unsupported on macOS")
```
JiriPavela, line 26: Let's create a new issue for these skips so that we track it somewhere and do not simply forget about it.

JiriPavela (approved):
Looks good to me, feel free to merge it.

perun/templates/macros/traces_selection.html.jinja2
robertvasek: You're right, I deleted the whole file. The data was not being used for generating the report but still probably shouldn't be visible in the repository.

PR #331, opened 2025-10-15
Fix CSV parsing in perun import

JiriPavela:
The new CSV parser in `perun import` now properly supports comments in CSV files, and gracefully handles cases where entries in the CSV file are missing or invalid.

### DIFF
--- a/perun/profile/imports.py
+++ b/perun/profile/imports.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 # Standard Imports
 from collections import defaultdict
-import csv
 from dataclasses import asdict, dataclass
 import gzip
 import json
@@ -12,9 +11,10 @@ import os
 from pathlib import Path
 import subprocess
 import sys
-from typing import Any
+from typing import Any, Sequence
 
 # Third-Party Imports
+import polars as pl
 
 # Perun Imports
 from perun import profile as profile
@@ -36,7 +36,7 @@ class _PerfProfileSpec:
     """
 
     path: Path
-    exit_code: int = 0
+    exit_code: int = -1
 
 
 @vcs_kit.lookup_minor_version
@@ -465,7 +465,7 @@ def _parse_perf_import_entries(
 
     where the CSV file is in the format
 
-      #Profile,Exit_code[,stat-header1]+
+      Profile,Exit_code[,stat-header1]+
       profile_path[,<exit code>[,<stat value>]+]
       ...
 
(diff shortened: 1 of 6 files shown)


perun/profile/imports.py
```diff
@@ -515,23 +515,30 @@ def _parse_perf_import_csv(
...
-        except StopIteration:
+            import_csv = pl.read_csv(csvfile_handle, comment_prefix="#")
+        except pl.exceptions.NoDataError:
             # Empty CSV file, skip
             log.warn(f"Empty import file {csv_path}. Skipping.")
             return
+        except pl.exceptions.ComputeError as e:
+            # The CSV file contains spurious columns, or is otherwise weirdly formatted.
+            # We log the issue so the user is aware of it, and we retry parsing it less strictly.
+            log.warn(
```
tfiedor, line 528: Can we get some hint for location of the error (I mean e.g. in Python you usually get nice helpul error message with `^` showing to exact column in the file)? Maybe some small context from the csv? (minor)

perun/profile/imports.py
```diff
@@ -551,25 +558,33 @@ def _parse_perf_entry(
...
+        log.warn(f"Invalid exit code '{entry[1]}' provided, using the default code {exit_code}.")
+    if exit_code != 0:
         log.warn("Importing a profile with non-zero exit code.")
+    profile_info = _PerfProfileSpec(profile_path, exit_code)
+
+    # Parse the stat values and add them to respective stats
+    for stat_value, stat_obj in zip(entry[2:], stats):
+        # Filter out missing values.
+        if stat_value is None:
+            continue
```
tfiedor: Maybe invert the condition, remove the continue and add the line 587.

perun/profile/stats.py
```diff
@@ -344,18 +345,27 @@ def from_values(cls, values: Iterable[float]) -> StatisticalSummary:
...
+        values = list(val for val in values if not math.isnan(val))
+        try:
+            quantiles = statistics.quantiles(values, n=20, method="inclusive")
+            return cls(
+                float(min(values)),
+                quantiles[2],  # p10
+                quantiles[5],  # p25
+                quantiles[10],  # p50
+                quantiles[15],  # p75
+                quantiles[18],  # p90
```
tfiedor, line 357: These indexes are quite counterintuitive. Is there a reason why `n = 20`? I assume that the quantiles are then mapped from (0, 100) to (0, 20). But it might make sense to refactor it a little.

tfiedor (approved):
I see nothing major. 

Just for interest: can you share some performance numbers of differences between the previous csv library and the polars? Is there improvement?

perun/profile/imports.py
JiriPavela: That would be difficult, since the parser does not report it directly, and we simply relay the (shortened) error message from polars to the user. To do that, we would likely need to write our own CSV parser and do the reporting manually. However, the typical input is expected to be quite small (usually up to 10 lines), so finding formatting issues should not be difficult.

perun/profile/imports.py
JiriPavela: Fixed.

perun/profile/stats.py
JiriPavela: Right, we need to obtain p10, p25, p50, p75, and p90, which means the highest common percentile multiple to compute the desired values is 5 (i.e., it suffices to compute p5, p10, p15, ..., p100). Computing all 100 percentiles is wasteful, as there are generally only few values and the resulting percentiles are interpolated. Any concrete ideas for refactoring this?

perun/profile/imports.py
tfiedor: So screw it then. If we can get something cheap, we should do it, but any complex changes are not worth the effort.

perun/profile/stats.py
tfiedor: What about something like
```suggestion
                quantiles[int(0.9 * n)],  # p90
```
It might look more complex, but it is clear this is 90th percentile of the sample `n` and if `n` changes we don't need to recompute it. WDYT?

tfiedor: Or we could extract all of this to constants:
```
SAMPLE_SIZE = 20
90_PERCENTILE = int(0.9 * SAMPLE_SIZE)
```

JiriPavela:
I did a very quick and crude measurements, and parsing an example workload similar to typical inputs seems to be about three times slower using polars -- however, the run time is dominated by the import times in both cases. So much so that measurements of import times and import + parsing are basically indistinguishable without detailed profiling.

On my machine, the polars version runs for approximately 0.095s (averaged over multiple runs) and the standard library version takes about 0.03s in user time. However, polars has better CSV-parsing functionality compared to the standard library (e.g., the ability to handle CSV comments) which was the primary reason for switching the libraries. I also plan to gradually introduce polars into the codebase as a replacement for pandas, and new import/showdiff versions will heavily leverage polars.

By the way, some performance comparisons between polars, pandas and other libraries can be found here: https://pola.rs/posts/benchmarks/.

PR #335, opened 2025-10-23
Feature/chatbot themed

robertvasek:
Chatbot dialog is fully themed based on selected theme.


robertvasek:
This merge request was setup by accident.

PR #336, opened 2025-11-03
Report; New monchromatic theme and chatbot theme adjustment

robertvasek:
Monochromatic theme which sets the whole report into shades of black and white was introduced.
Also chatbot has adjusted colors which match the state of current theme.
Report with new features can be viewed here: 
[report_x.html](https://github.com/user-attachments/files/23304446/report_x.html)


JiriPavela (approved):
Very nice job. I noticed a small issue with the color styling of the Links menu, which seems to have some of the colors inverted, but we can fix it in a standalone future PR. I think we can merge this.

PR #343, opened 2026-01-11
Add new report-folded diffview

### LINKED ISSUE #305: Inefficient processing of profiling data in `perun showdiff report`

JiriPavela:
This PR splits the `showdiff report` command into two separate variants: `native` and `folded`. The _native_ variant processes already imported or collected Perun profiles, whereas the _folded_ variant works on external folded profiles (e.g., perf-folded or ebpf-folded). The report can be further extended to support even more variants of external profiles, e.g., ftrace or callgrind. The `report` command hierarchy has been reworked as follows: `perun showdiff [common showdiff options] report [common report options] native|folded [variant-specific arguments and options]`.

The folded variant has been implemented from scratch using the [Polars](https://pola.rs/) library to improve the performance of data processing. Moreover, the folded variant implements both serial and parallel creation of flamegraph grids (can be selected through CLI). Especially for large inputs, the parallel creation can lead to about 2x speedup at the cost of additional memory consumption by all the flamegraph scripts running in parallel.

Additionally, the `folded` report also computes additional difference metrics and identifies the most expensive functions and traces w.r.t. multiple metrics. These results are currently not reflected in the generated HTML, however, they will soon be. 

This PR resolves #305.

**Performance**

Crucially, the new report variant shows significant performance improvements over the native approach on both small and large inputs. I have measured the performance on my laptop using the `stress-ng mmapmany` and `fio` workloads as representatives of small and large inputs:

| Input | Profile Size (MB) |
|--------- | ----------------------- |
| `mmapmany baseline` | ~ 0.26 |
| `mmapmany target`  | ~ 0.71 |
| `fio baseline`  | ~ 158 |
| `fio target`   | ~ 74 |

For small inputs, the measured improvements are not as significant, but still notable: approximately **2.4x lower memory requirements** (compared to the most memory-hungry `import` command) and approximately **5x speedup** in total (a single `report folded` command as opposed to a pipeline of 2 `import`s and `report native` commands).

| Perun Command | Peak Memory Usage (MB) | Elapsed Time (s) |
| ------------------------- | ------------------------------------- | ----------------------- |
| `import baseline` | ~ 120  | ~ 1.5 |
| `import target`     | ~ 160  | ~ 2 |
| `report native`      | ~ 104  | ~ 3.5 |
| `report folded serial`   | ~ 67 | ~ 1.45 |
| `report folded parallel` | ~ 67 * | ~ 1.3 |

(* plus additional memory consumed by the 4 flame graph processes running in parallel in the background.)

For large inputs, the performance improvements are overwhelming: up to **19x lower memory requirements** (for serial variant) and **12x to 24x speedup** of serial, resp. parallel, version.

| Perun Command | Peak Memory Usage (GB) | Elapsed Time (s) |
| ------------------------- | ------------------------------------- | ----------------------- |
| `import baseline` | ~ 15  | ~ 160 |
| `import target`     | ~ 6.2  | ~ 70 |
| `report native`      | ~ 6.5 | ~ 500 |
| `report folded serial`   | ~ 0.8 | ~ 56 |
| `report folded parallel` | ~ 0.8* | ~ 30 +- 2 |

(* plus additional memory consumed by the 4 flame graph processes running in parallel in the background.)

Additionally, the `folded` report also saves disk space as it does not store any perun-native profiles.

**Future Work** 

- There are currently patches of duplicate or very similar code between `report native`, `import`, and `report folded` that will be refactored later (see #345).

- The current performance bottlenecks are the `difffolded.pl` and `flamegraph.pl` scripts (coupled with the need to store the postprocessed folded data back into files) that have to needlessly recompute a lot of data that we already computed in our code. A rewrite of the perl scripts into native Python should bring further significant time and memory savings (see #344). 

- A lot of unnecessary memory is consumed by representing the input profiles in polars DataFrames. Even though much more efficient data structures exist, it proved to be difficult to implement them efficiently in Python. I plan to try again once more. Such change could also improve the precision of our recursion squashing post-processing (see #345).


perun/cli_groups/showdiff_cli.py
```diff
@@ -170,10 +230,14 @@ def showdiff_group(_: click.Context, **__: Any) -> None:
...
 
         perun showdiff -m HEAD~1 0@i 1@i report
+
+    Refer to concrete commands for details on their difference visualizations and supported profile
+    formats.
     """
 
 
 @showdiff_group.command()
+@from_perun_profiles
```
tfiedor: For these click option groups, I would name them in a way that shows it is option group, so maybe `from_perun_profiles_options`?

perun/view_diff/report_folded.py
```diff
@@ -0,0 +1,2259 @@
...
+        """Compute the minimum width threshold for flamegraph blocks to be displayed.
+
+        The threshold is needed to correctly compute the maxtrace value so that flamegraphs in
+        a grid are correctly aligned and have the appropriate height.
+
+        Reconstructed from the flamegraph.pl script.
+
+        :return: the minimum width threshold
+        """
+        try:
```
tfiedor, line 646: Can the helper function for try_division be used here?

perun/view_diff/report_folded.py
```diff
@@ -0,0 +1,2259 @@
...
+        tabular_profile.traces.group_by("func", maintain_order=True)
+    ):
+        # Obtain a row corresponding to the function from the function profile. We can access it
+        # using an index thanks to the profile being sorted.
+        row = df_row(tabular_funcs, idx)
+        # Determine the state of the function. If it is not in the set of common functions, then we
+        # look at the 'inclusive' and 'inclusive_target' values to determine wheter it is
+        # a baseline-only or target-only function.
+        state: Literal["not in baseline", "not in target", "in both"] = "in both"
+        if row[0] not in common_funcs:
```
tfiedor, line 1611: I would probably comment the indexes, as 0, 1, 3 seems magical, and one would think, if there might have been `row[2]` instead of `row[3]`. (Similar to below on 1624+ lines)

tfiedor (approved):
LGTM, I assume the `report_folded.py` will be refactored in further PRs as there is quite a lot of TODOs and FIXMEs (I skimmed through that file).

perun/cli_groups/showdiff_cli.py
JiriPavela: Good call, I renamed it to `perun_profile_list_options`.

perun/view_diff/report_folded.py
JiriPavela: Good point, I added comments there as well.

perun/view_diff/report_folded.py
JiriPavela: I suppose you meant the `safe_division` function? It is not applicable here, really. We need to set the threshold to `0.0` if the division fails to disable traces filtering, whereas the `safe_division` would return a very large number that would result in filtering all the traces, hence computing an incorrect `max_trace` value.

JiriPavela:
> LGTM, I assume the `report_folded.py` will be refactored in further PRs as there is quite a lot of TODOs and FIXMEs (I skimmed through that file).

Correct, `showdiff flamegraph`, `showdiff report native`, and `showdiff report folded` will all be refactored once I optimize the `folded` variant even more.

PR #348, opened 2026-01-21
Resolve minor issues in the recent 0.26.0 version 

JiriPavela:
This PR fixes some issues in the recent version and implements some minor changes:

- The nested traces table in `showdiff report` now has correctly styled table headers. All table headers now also use the same column names and tooltips.
- Perun now supports both long and short help options (`--help` and `-h`) for CLI commands.
- The `showdiff report`, `showdiff report native`, and `showdiff flamegraph` help docstrings have been changed to be more user-friendly.
- The integrated chatbot has been updated to the latest version.


perun/templates/scripts/value_formatters_js.html.jinja2
```diff
@@ -196,15 +196,21 @@
             result += '<td style="white-space: pre-wrap">' + value[5] + "</td>";
             result += "</tr>";
         });
+
```
tfiedor, line 199: This looks weird. Why is this here?

tfiedor (approved):
LGTM

PR #350, opened 2026-02-04
New traces table introduced

robertvasek:
Dependency on library `dataTables` was removed
- instead own custom logic of traces table was introduced
- most of the functionality was retained
- this gives us more flexibility in extending the traces table and shaping it according to our own requirements


perun/templates/scripts/custom_table.js
```diff
@@ -0,0 +1,476 @@
...
+        if (absNum >= 1.0e9) {
+            return (num / 1.0e9).toFixed(3) + " G";
+        } else if (absNum >= 1.0e6) {
+            return (num / 1.0e6).toFixed(3) + " M";
+        } else if (absNum >= 1.0e3) {
+            return (num / 1.0e3).toFixed(3) + " K";
+        } else {
+            return num.toString();
+        }
+    }
```
JiriPavela, line 41: The same function seems to be in the `custom_table.html.jinja2` file as well. Do we need it defined in two files separately for some reason?

JiriPavela (approved):
Good job, I would consider renaming the `custom_table` files to something more descriptive.

PR #352, opened 2026-03-22
FEAT: Traces detail (popup) in Traces table section

robertvasek:
This PR introduces new popup which is part of traces table section.
- This popup shows details about particular trace
  - shows detailed info which is also visible in parent's table row
  - additionaly shows each particular trace which is part of the given trace (always max 10) 
-  Values respect 3-digit formatting -> 0.33M, 3.34 M, 33.4 M, 334M
- Other visual enhancements introduces as well in the traces table including tooltips, sortings, info icons with tooltips
- Refactorings of some legacy code which were previously used in datatables library


robertvasek:
Test will probably not go through since there is a bug in tests which we discussed with @JiriPavela earlier.

robertvasek:
Codacy had issue with some variables because it cannot see that the variables are being used in jinja2 templates. Comments were introduced in order to fix this issue.

JiriPavela (approved):
Great, I think we can merge it. The failing tests should work after the merge.

PR #356, opened 2026-03-30
Add a stub overview section to folded report

JiriPavela:
The overview section will list top baseline-only, target-only, and common traces and functions w.r.t some diff metrics. This PR adds the propagation of the computed results into the HTML report so that the frontend can read the data.

### DIFF
--- a/perun/view_diff/report_folded.py
+++ b/perun/view_diff/report_folded.py
@@ -457,6 +457,10 @@ class KeyDiffs:
     The class stores the top increases and decreases of a key (metric) for baseline-only,
     target-only, and common partitions of a merged profile.
 
+    The individual DataFrames contain only two columns: "symbol" and "diff".
+     - "symbol" identifies either a trace or a function by their IDs, and
+     - "diff" is the value of the diff metric.
+
     :ivar baseline_top_inc: the largest increases in baseline-only records
     :ivar baseline_top_dec: the largest decreases in baseline-only records
     :ivar target_top_inc: the largest increases in target-only records
@@ -1171,9 +1175,10 @@ def generate_report_from_folded(
         notes_enabled=True,
         links=report_links,
         default_theme=cli_kwargs.get("default_theme", "dark"),
-        # FIXME: the top diffs will be used in the future.
-        top_trace_diffs=trace_top_diffs,
-        top_func_diffs=func_top_diffs,
+        top_diffs_trace_inclusive=iterate_top_diffs(trace_top_diffs[0]),
+        top_diffs_trace_exclusive=iterate_top_diffs(trace_top_diffs[1]),
+        top_diffs_func_inclusive=iterate_top_diffs(func_top_diffs[0]),
+        top_diffs_func_exclusive=iterate_top_diffs(func_top_diffs[1]),
     )
     log.minor_success("HTML report", "rendered")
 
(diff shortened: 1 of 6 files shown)


perun/view_diff/report_folded.py
```diff
@@ -1472,18 +1477,20 @@ def compute_top_diffs(
...
-            common_lf.bottom_k(records_num, by=inclusive_diff_key),
-            baseline_only_lf.top_k(records_num, by=exclusive_diff_key),
-            baseline_only_lf.bottom_k(records_num, by=exclusive_diff_key),
-            target_only_lf.top_k(records_num, by=exclusive_diff_key),
-            target_only_lf.bottom_k(records_num, by=exclusive_diff_key),
-            common_lf.top_k(records_num, by=exclusive_diff_key),
-            common_lf.bottom_k(records_num, by=exclusive_diff_key),
+            # Inclusive diff metric.
+            find_top_diffs(baseline_only_lf, records_num, symbol_key, inclusive_diff_key),
+            find_top_diffs(baseline_only_lf, records_num, symbol_key, inclusive_diff_key, False),
```
tfiedor: I would suggest to use keyword for the `False` so it is better readable, since this way one does not know immediately  what is the parameter there), so `top=False`.

perun/view_diff/report_folded.py
```diff
@@ -2259,3 +2286,29 @@ def _update_profile_features(
     # we need to recompute the total number of consumed resources.
     if profiles_count > 1 and aggregation_func != pl.sum:
         features.total_resources = int(prof["exclusive"].sum())
+
+
+def find_top_diffs(
+    data_lf: pl.LazyFrame, count: int, symbol_key: str, diff_key: str, top: bool = True
```
tfiedor: I kindof like using `*` as follows:
```suggestion
    data_lf: pl.LazyFrame, count: int, symbol_key: str, diff_key: str, *, top: bool = True
```
That forces the `top` to be used as keyword argument always. When you have many parameters or booleans I think it is godsend.

perun/view_diff/report_folded.py
```diff
@@ -2259,3 +2286,29 @@ def _update_profile_features(
...
+
+    :param data_lf: the data where to search for the maximal/minimal records.
+    :param count: the upper limit on the number of top records to search for.
+    :param symbol_key: the name of the column with symbols (e.g., "trace")
+    :param diff_key: the name of the column containing the diff metric
+    :param top: specifies whether to search for the maximal or minimal records.
+
+    :return: a lazy frame containing the maximal/minimal records.
+    """
+    search_func = pl.LazyFrame.top_k if top else pl.LazyFrame.bottom_k
```
tfiedor: It would make sense to also switch from boolean parameter to supply the function; but looks like we have two uses cases anyway so it is fine.

tfiedor (approved):
LGTM, made few suggestions that might help readability.

Also it might make sense to invest (you can try claude/codex/whatever) some effort to pack the whitespaces in the jinja template (in my experience, it can save quite some bytes in these reports).

perun/view_diff/report_folded.py
JiriPavela: Good idea, fixed.

perun/view_diff/report_folded.py
JiriPavela: Fixed.

perun/view_diff/report_folded.py
JiriPavela: This would normally be the preferred option to a bool parameter, but the `top_k` and `bottom_k` polars functions have quite complex types, which would make the type hints of this little helper function too cumbersome.

JiriPavela:
> LGTM, made few suggestions that might help readability.
> 
> Also it might make sense to invest (you can try claude/codex/whatever) some effort to pack the whitespaces in the jinja template (in my experience, it can save quite some bytes in these reports).

This is a good suggestion. Rather than just removing the whitespace, I think we should instead try to use something similar to JS minification, which has the potential to save even more space.

PR #364, opened 2026-05-04
Provide native Python implementations of Flame Graph scripts

### LINKED ISSUE #344: Reimplement the `flamegraph.pl` and `difffolded.pl` in Python
The `flamegraph.pl` and `difffolded.pl` scripts have generally an okay performance, however, creating a flamegraph grid (2x2) for large profiles often consumes too much time and memory. We should re-implement both scripts in native Python and implement some performance optimizations when we already have the profile data pre-processed (as is the case in the `showdiff` command.

JiriPavela:
# Overview

This PR adds Python translations of the original Perl scripts `difffolded.pl` and `flamegraph.pl`. The Python versions offer better performance than the original Perl versions and are easier to integrate directly into Python code if needed.

It also adds a helper script that combines both `difffolded.py` and `flamegraph.py` to draw a differential flame graph directly from two profiles. For large profiles, this is even faster than piping the two python scripts together. 

Finally, the Python versions extend the original Perl scripts with new CLI options and fixes some bugs, e.g., missing frames in differential flame graphs.

The Python scripts are now used by default when generating folded reports. It is possible to revert to the Perl versions via a flag `--flamegraph-use-perl-scripts`.

# Performance measurements

Similarly to #343, I did a crude local performance measurements using profiles from two `stress-ng` benchmarks, `mmapmany` and `fio`, as representatives of small and large inputs. I have measured the performance of both the individual scripts and generating a folded report.

| Input | Profile Size (MB) |
|--------- | ----------------------- |
| `mmapmany-base` | ~ 0.26 |
| `mmapmany-tar`  | ~ 0.71 |
| `fio-base`  | ~ 158 |
| `fio-tar`   | ~ 74 |

**Summary**

Overall, the Python scripts are slower and consume more memory for small folded profiles. This is caused mostly by the Python VM overhead, which accounts for ~40-50ms startup time and ~3.7 MB heap memory usage. However, this additional cost is mostly insignificant when constructing reports.

For large profiles, the new Python versions show significant improvements both in execution time and memory consumption. For example, generating a `fio` folded report shows a 2x speedup.

Despite the lackluster performance for small profiles, the Python version will help us integrate flame graph generation directly into our code (instead of creating `subprocess`es and passing data through files), which will lead to performance gains in the future. It will also make it easier to extend the flame graphs with new features that had to be written in Perl before.

**Individual scripts**

| Command | Peak Memory Usage (MB) | Elapsed Time (s) |
| ------------------------- | ------------------------------------- | ----------------------- |
| `difffolded.py fio-base fio-tar` | 114  | 1.33 |
| `difffolded.pl fio-base fio-tar`  | 334  | 2.8 |
| `difffolded.py mmapmany-base mmapmany-tar` | 4.4  | 0.067 |
| `difffolded.pl mmapmany-base mmapmany-tar`  | 1.5  | 0.014 |
| `flamegraph.py fio-base` | 101  | 1.38 |
| `flamegraph.pl fio-base`  | 544  | 10.8 |
| `flamegraph.py mmapmany-base` | 7.7  | 0.087 |
| `flamegraph.pl mmapmany-base`  | 3.7  | 0.051 |
| `difffolded.py fio-base fio-tar \| flamegraph.py` | 114 + 101  | 2.44 |
| `difffolded.pl fio-base fio-tar \| flamegraph.pl`  | 334 + 593  | 11.82 |
| `diff_flamegraph.py fio-base fio-tar` | 114  | 1.99 |
| `difffolded.py mmapmany-base mmapmany-tar \| flamegraph.py` | 4.4 + 7.7  | 0.104 |
| `difffolded.pl mmapmany-base mmapmany-tar \| flamegraph.pl`  | 1.5 + 3.9  | 0.049 |
| `diff_flamegraph.py mmapmany-base mmapmany-tar` | 8.5 | 0.114 |

**Folded Reports**

| Command |  Elapsed Time (s) |
| ----------------| ------------------------ |
| `report --parallel --python fio-base fio-tar` | 18.1 |
| `report --parallel --perl fio-base fio-tar` | 33.4 |
| `report  --serial --python fio-base fio-tar`  | 24.2 |
| `report  --serial --perl fio-base fio-tar`   | 63.2 |
| `report --parallel --python mmapmany-base mmapmany-tar` | 1.18 |
| `report --parallel --perl mmapmany-base mmapmany-tar`   | 1.1 |
| `report  --serial --python mmapmany-base mmapmany-tar`   | 1.4 |
| `report  --serial --perl mmapmany-base mmapmany-tar`   | 1.2 |

This PR resolves #344.

### DIFF
--- a/perun/scripts/difffolded.py
+++ b/perun/scripts/difffolded.py
@@ -245,12 +245,17 @@ def diff_folded_lazy(
         ...
     ```
 
+    The lazy variant will store the baseline profile in memory and then, when
+    possible, will yield diff records on demand without storing the entire
+    target profile in memory.
+
     Note that the most efficient use-case for lazy diff is when neither sorting
     nor normalization is requested. Otherwise, requesting normalization and/or
     sorting will lead to slightly slower execution and more memory consumption
-    as the entire target profile needs to be parsed eagerly. Still, minor
-    performance gains compared to a fully eager approach can be expected as the
-    normalization and sorting can be done at least partially lazily.
+    as the entire target profile needs to be eagerly parsed and stored in
+    memory. Still, minor performance gains compared to a fully eager approach
+    can be expected as the normalization and sorting can be done at least
+    partially lazily.
 
 
     :param baseline: a path to the baseline folded profile.
--- a/perun/scripts/difffolded.py
+++ b/perun/scripts/difffolded.py
@@ -1,0 +1,667 @@
+#!/usr/bin/env python3
+
+"""difffolded.py 	Diff baseline and target folded stack files.
+                    Use this for generating flame graph differentials.
+
+USAGE: ./difffolded.py [-hnst] [-f integer|float|auto] baseline target | ./flamegraph.py > diff2.svg
+
+Options are described in the usage message (-h).
+
+The flamegraph will be colored based on higher samples (red) and smaller
+samples (blue). The frame widths will be based on the target (2nd folded file).
+This might be confusing if stack frames disappear entirely; it will make
+the most sense to ALSO create a differential based on the baseline widths,
+while switching the hues; eg:
+
+ ./difffolded.py target baseline | ./flamegraph.py --negate > diff1.svg
+
+Here's what they mean when comparing a before and after profile:
+
+diff1.svg: widths show the before profile, colored by what WILL happen
+diff2.svg: widths show the after profile, colored by what DID happen
+
+INPUT: See stackcollapse* programs.
+
+OUTPUT: The full list of stacks, with two columns, one from each file.
+If a stack wasn't present in a file, the column value is zero.
+
+folded_stack_trace count_from_baseline count_from_target
+
+eg:
+
+funca;funcb;funcc 31 33
+...
+
+
+*******************************************************************************
+* Python version details
+*******************************************************************************
+
+This Python version is meant as a drop-in replacement for the original Perl
+script. As such, it requires only a Python interpreter with the Python standard
+library: it deliberately avoids any 3rd party dependencies.
+
+The Python version has been optimized for processing large folded profiles and,
+based on some crude local measurements (Python 3.14, cached bytecode), appears
+to be almost 2x faster for an example ~150 MB folded profile. Moreover, even
+more speedup (~3x) can be achieved by importing the script and using its (lazy)
+API directly. Small profiles tend to exhibit a slowdown compared to the Perl
+version as the execution time is dominated by the Python interpreter startup
+cost (~40ms+ on a local machine). Also note that the optimizations cause some
+code duplication as we have no macros or templates in Python.
+
+Additionally, this version introduces some minor changes compared to the
+original Perl script:
+
+ - Unhandled zero division in normalization is now detected, a warning is
+   issued to the user, and the script falls back to the non-normalized output.
+   Additionally, the exact float comparison has been replaced with a more
+   reliable comparison.
+
+ - Scientific notation is now supported both on the input and output.
+
+ - The script can be used either through a CLI exactly as the Perl version, or
+   through a programmatic Python API by importing the file and calling either
+   the 'diff_folded' or 'diff_folded_lazy' functions. Using the functions
+   directly will lead to an even bigger speedup (crude measurements show up to
+   3x speedup).
+
+ - The CLI has a new '--formatter' option related to the standard output.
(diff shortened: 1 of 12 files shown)


perun/scripts/difffolded.py
```diff
@@ -0,0 +1,668 @@
...
+    ) -> None:
+        self.data: Iterable[tuple[str, float, float]] = data_generator
+        self.baseline_sum: float = baseline_sum
+        self.target_sum: float = target_sum
+
+
+def diff_folded(
+    baseline: str,
+    target: str,
+    *,
```
tfiedor, line 186: Did I impart wisdom on you? (I have some memory, possibly false, of telling you about this neat hack; not a comment, just wondering if I'm hallucinating).

perun/scripts/difffolded.py
```diff
@@ -0,0 +1,668 @@
...
+def diff_folded_lazy(
+    baseline: str,
+    target: str,
+    *,
+    striphex: bool = False,
+    normalize: bool = False,
+    sort_stacks: bool = False,
+    **_: Any,
+) -> FoldedDiffLazy:
+    """Diff two folded profiles lazily.
```
tfiedor, line 237: Let's briefly document what "Lazy" means

perun/scripts/difffolded.py
```diff
@@ -0,0 +1,668 @@
...
+    try:
+        with open(file_path, encoding="utf-8") as file_handle:
+            if hex_strip_pattern is None:
+                for line in file_handle:
+                    stack, count_str = str_rsplit(line, maxsplit=1)
+                    try:
+                        count = float(count_str)
+                        total_baseline += count
+                        baseline_data[stack] = count
+                    except ValueError:
```
tfiedor, line 363: I believe there is contexlib function for this usecase.

perun/scripts/difffolded.py
```diff
@@ -0,0 +1,668 @@
...
+    scale: float | None = compute_scale_factor(baseline_sum, target_sum)
+    if scale is not None:
+        return (
+            (stack, count_baseline * scale, count_target)
+            for stack, count_baseline, count_target in data_iter
+        )
+    return data_iter
+
+
+def build_strip_hex_pattern(striphex: bool) -> re.Pattern[str] | None:
```
tfiedor, line 540: Consider caching? In functools there should be decorator for this.

tfiedor (approved):
Impresive! Note that this might be interesting to other devs (though I think there is implementation of flamegraphs in rust right? Inferno it is called). Might be worth writing some blog pointing to this implementation and add it to hackernews.

I went through everything but the main `flamegraph.py` (those 3k are too much for my free time). I trust it is as good as rest.

For more optimization (so I offer at least some wisdom), perhaps adding caching from functools (`cache`, `lru_cache`) could give us more.

perun/scripts/difffolded.py
JiriPavela: Yes, that is correct.

perun/scripts/difffolded.py
JiriPavela: Fixed. I added a new paragraph to the docstring.

perun/scripts/difffolded.py
JiriPavela: Do you mean the [`suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) context manager? That would in fact cause a significant regression (about 30% on larger profiles based on my measurement) as the manager object needs to be repeatedly created in each loop iteration.

tfiedor: Good point!

perun/scripts/difffolded.py
JiriPavela: This function is called at most once in each run, as it pre-compiles a regex pattern that is then used repeatedly. However, this brings me to an interesting observation: the pattern is very simple and fixed (i.e., not user-supplied); as such, we could replace it with native code and get rid of the regex library completely. I will add this to my TODO list before turning this into a public repo.

JiriPavela:
Great, thanks for the review. I believe we already discussed this in private, but I indeed plan to eventually release this as a new repo / library. I have many ideas that, I believe, could make this library stand out among other flame graph repositories and libraries (I am aware of Inferno and https://github.com/flamegraph-rs/flamegraph). However, there is still much work that needs to be done before I feel comfortable releasing it for public use.

PR #365, opened 2026-05-06
Add new sub-root frame to flame graph grids

JiriPavela:
Our flame graph grid scales all graphs using a common root node with a fixed width. While this is quite helpful in assessing the difference between the total consumption of both profiles, it may also cause issues with inspecting the individual profiles when the difference is too big.

Concretely, it is not possible to zoom all frames in one profile to their unscaled width and visually compare them with the other profile.

Screenshot From 2026-05-06 16-47-32

\
This commit attempts to solve this issue by adding an extra sub-root node to our grid. This sub-root is added to a flame graph only when the `--total` parameter is specified, and it allows the user to zoom all frames of the profile to their unscaled widths.

Screenshot From 2026-05-06 16-46-23

\
Screenshot From 2026-05-06 16-46-34

\
Furthermore, the `--minwidth` threshold is now calculated using the actual total consumption (the sum of all counts in the folded profile) instead of the --total parameter. This aims to prevent excessive filtering when the two profiles have drastically different total consumptions.

### DIFF
--- a/perun/scripts/diff_flamegraph.py
+++ b/perun/scripts/diff_flamegraph.py
@@ -42,8 +42,9 @@ def create_diff_flame_graph(
     notes: str = ...,
     outfile: None = ...,
     random: bool = ...,
-    rootnode: str = ...,
     reverse: bool = ...,
+    rootnode: str = ...,
+    subrootnode: str = ...,
     subtitle: str = ...,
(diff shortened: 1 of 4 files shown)


tfiedor (approved):
LGTM

PR #366, opened 2026-05-13
Add exclusive consumption tracking to flame graphs

JiriPavela:
This PR adds support for displaying exclusive (self) resource consumption in report flame graphs.

The users may now switch between inclusive and exclusive views per each flame graph or globally for all flame graphs. The inclusive view for non-diff graphs corresponds to the default flame graph generated by flamegraph.pl. The exclusive view colors the original frames in white and adds colored overlay frames with their width scaled according to the exclusive consumption.
    
By default, diff flame graphs showed differences in exclusive consumption. This was especially unintuitive when comparing diff (exclusive) and non-diff (inclusive) graphs side-by-side. Hence, we now also compute inclusive consumption deltas and show these as the default view to stay consistent with non-diff graphs. In diff graphs, the exclusive consumption overlay frames are not width-scaled, but instead change their color hue.

Moreover, the order of the diff flame graphs has been switched so that vertically aligned graphs have identical structure which makes it easier to map the changes. Lastly, the tooltips and text descriptions now show both inclusive and exclusive consumptions.

- Since the flame graph JS code is becoming hard to maintain (and is generally slow for large graphs), I opened a new issue #367  

**Original State** (diff graphs are colored based on exclusive consumption deltas)

Screenshot From 2026-05-13 16-02-31

\
**New Inclusive View**
Screenshot From 2026-05-13 16-27-33

\
**New Exclusive View**
Screenshot From 2026-05-13 16-27-58

\
**Updated Descriptions**
Screenshot From 2026-05-13 16-29-18


tfiedor (approved):
LGTM

PR #368, opened 2026-05-18
Extend the Tabular View with a table containing exclusive consumption data

JiriPavela:
This commit renames the `Traces Table` section to `Tabular View` and extends the original table (with inclusive consumption data only) with a second table with exclusive consumption records. Furthermore, the trace details can now show both top inclusive and top exclusive traces for a given function.

Adding a second separate table instead of merging the columns into a single table was an intentional design decision: this way the user can easily switch between two separate tabular views with possibly different filtering or sorting for inclusive and exclusive consumption data.

**Inclusive consumption table + detail**

Screenshot From 2026-05-18 22-21-43

\
Screenshot From 2026-05-18 22-22-46

\
**Exclusive consumption table + detail**

Screenshot From 2026-05-18 22-22-01

\
Screenshot From 2026-05-18 22-23-18


tfiedor (approved):
LGTM

